PR #19, opened 2022-01-27
Add tests for lattice object

kessler-frost:
Unit tests to be written for following functions:
- [x] `init`
- [x] `build_graph`
- [x] `draw`
- [x] `__call__`
- [ ] `check_constraint_specific_sum`
- [ ] `check_consumable`
~`dispatch`~
~`dispatch_sync`~
~`_server_dispatch`~
- [x] `lattice`


kessler-frost:
Also check time taken by `build_graph`.

wjcunningham7:
@kessler-frost I updated this to be compatible with the latest version of Covalent, added an init test, and removed tests related to dispatching, which are already handled in `dispatcher_plugins/local_test.py` given that we changed how dispatches can be done.  I'm going to attempt to continue working on tests here for practice.

wjcunningham7:
another important note -- check how I changed the imports in `covalent._workflow.__init__.py`.  This is because we cannot mock any function inside `lattice.py` or `electron.py` since the original import structure meant `covalent._workflow.electron`, for instance, would refer to a function and not the module.  This small change fixes that.

wjcunningham7:
@FyzHsn there are a variety of unused or deprecated functions which may artificially make coverage appear lower:
- `draw_inline`
- `check_constraint_specific_sum`
- `check_consumable`
- `dispatch`
- `dispatch_sync`

How do we prevent deprecated functions from lowering code coverage?

FyzHsn (approved):

PR #27, opened 2022-01-28
Support for ARM

scottwn:
Adding support for ARM via the following changes.

- Installing `gcc` in Docker image
- Updating versions for `matplotlib` and `dask` in `requirements.txt` to current versions that have better ARM support
- Using `docker buildx` in `publish` workflow to build images for multiple architectures and push manifest to ecr


wjcunningham7 (approved):
looks good, confirmed the pipeline runs properly; did not test the upload but I assume that won't be a problem

Issue #50, opened 2022-01-30
Tests for PyPi releases


kessler-frost:
### What should we add?

As we saw recently that even though things were working as expected when covalent is installed using the GitHub repo, if someone did `pip install cova` they wouldn't have been able to use covalent. We should add tests/actions in `pypi.yml` to check whether installing from pip works as intended before every merge to develop or master.

### Describe alternatives you've considered.

For starters we can add a diff check between what is built by `python setup.py sdist` and what is there in the repo barring files/dirs containing docs, tests, tutorials, etc.

PR #53, opened 2022-01-30
Fix for UI server not showing when installed through pip

kessler-frost:

When covalent is installed through `pip install cova` the UI server which is supposed to run on `localhost:47007` is not run due to the missing `webapp` folder in that package. This PR will add that folder to `setup.py` so that it can be packaged properly.

@wjcunningham7 should I also try to resolve issue #50 in this PR?


wjcunningham7:
@kessler-frost yes also try to fix #50

wjcunningham7:
evidence of the pypi validation test passing: https://github.com/AgnostiqHQ/covalent/runs/5012012294?check_suite_focus=true

valkostadinov (dismissed):

scottwn (approved):

PR #55, opened 2022-01-31
Switch from lattice.dispatch to ct.dispatch

### LINKED ISSUE #48: Switch from `workflow.dispatch` to `covalent.dispatch`
### What should we add?

Currently we are using `workflow.dispatch` to send a lattice `workflow` for execution to the dispatcher server. Maybe we can use a more intuitive approach for dispatching which is similar to how we obtain the result for a particular dispatch. The proposal is to have a function `covalent.dispatch` similar to `covalent.get_result` so that it is more intuitive for users to send a job the same way they get the result.

Resolving this issue may also involve updating the `README.md`s, tutorials, documentation, etc. So it will turn into a big issue.

_No response_

kessler-frost:

Resolves #48


kessler-frost:
Only thing left to update is the website. `rst` files, readme, and contributing has also been updated.

kessler-frost:
Will need to run the how tos, tutorials again to check if they are still working. Tests seem to be passing which take how-tos into account.

santoshkumarradha:
> The only place its not updated is the website.

@kessler-frost  can you slack me once this is merged and in pip ? I will update it there.

santoshkumarradha (changes requested):
Great job @kessler-frost  !  🔥 

```python
    dispatcher=ct.dispatch(run_experiment)
    id=dispatcher(*params)
```

Only suggestion - 

Using nested functions is an advanced pattern and not a beginner to intermediate  friendly pattern. Could you change user facing example (atleast in read me) to a pattern as above where it is a bit more easier to digest than nested calls like `func()()` ?

doc/source/api/api.rst
```diff
@@ -36,6 +37,23 @@ Lattice
...
+""""""""""""""
+
+Dispatching jobs to the dispatcher servers
+
+.. autofunction:: dispatch
+.. autofunction:: dispatch_sync
+
+
+.. autoclass:: covalent._dispatcher_plugins.LocalDispatcher
+    :members:
```
santoshkumarradha, line 52: can you hide private members ?

covalent/_dispatcher_plugins/local.py
```diff
@@ -0,0 +1,72 @@
...
+
+from .._results_manager.result import Result
+from .._results_manager.results_manager import get_result
+from .._workflow.lattice import Lattice
+from .base import BaseDispatcher
+
+
+class LocalDispatcher(BaseDispatcher):
+    @staticmethod
+    def dispatch(orig_lattice: Lattice) -> str:
```
santoshkumarradha, line 36: We would ned docstrings here for it to show up in API part. And also the the dispatcher properties like port etc.. ?

kessler-frost: Added the dispatcher address and docstrings.

covalent/_workflow/lattice.py
```diff
@@ -287,90 +285,13 @@ def check_consumable(self) -> None:
                     )
                 )
 
-    def dispatch(self, *args, **kwargs) -> str:
```
santoshkumarradha, line 290: @kessler-frost I forgot to add this, can you  have a pass through of this calling `ct.dispatch` with a depreciation warning that this will be removed in few releases?

santoshkumarradha: or a meaning full error with a pass passthrough without dispatch.

tests/functional_tests/serialization_test.py
```diff
@@ -38,10 +38,7 @@ def electron_function(x):
 
 
 @etron
```
FyzHsn, line 40: This looks like a typo

tests/functional_tests/serialization_test.py
```diff
@@ -38,10 +38,7 @@ def electron_function(x):
...
-@cova.lattice(
-    results_dir="./",
-    dispatcher="super long fake dispatcher to test really long arguments                                                                                                                                                                                end_fake_dispatcher_name",
-)
+@cova.lattice(results_dir="./")
```
FyzHsn, line 41: I'm little confused about import covalent and then covalent again as cova.

FyzHsn (changes requested):
I have added a couple of (non-blocking) questions that require some clarification. Also, for the sake of good practise, is it possible to add some unit tests for base and local.py? I know this is a huge issue already.

FyzHsn:
@kessler-frost Looks great! Just had a couple of questions and I think it would be good practise to write some unit tests for base and local.py.

tests/functional_tests/serialization_test.py
kessler-frost: I think these changes were introduced in a separate PR done before I started working on this PR. As for the reason, I think @jkanem did this in order to test that things still work if someone imports `covalent` as `cova`.

tests/functional_tests/serialization_test.py
kessler-frost: This has the same reason as mentioned below, its for a test (I think related to importing) and was introduced before this PR.

kessler-frost:
@FyzHsn I agree, I'll add some tests for `BaseDispatcher` and `LocalDispatcher`.

FyzHsn:
> @FyzHsn I agree, I'll add some tests for `BaseDispatcher` and `LocalDispatcher`.

I really don't want to be the tests guy. But somehow, I can never escape it.

CONTRIBUTING.md
```diff
@@ -220,7 +220,7 @@ fi
 -   For single line statements and associated output, use Python console syntax (pycon):
 
 ```python
->>> pipeline.dispatch(**params) # Dispatching a workflow returns a unique dispatch id.
+>>> ct.dispatch(pipeline)(**params) # Dispatching a workflow returns a unique dispatch id.
```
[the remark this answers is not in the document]
kessler-frost: If we go down the `ct.dispatch(pipeline, pipeline_args)` path one major issue is that, say the user wants to pass some positional arguments and some keyword arguments, how can he do that? `ct.dispatch(pipeline, pipeline_args = {1, x: 3, y: 4}` doesn't work as `1` doesn't have a key. Then they'll have to pass positional arguments as different than keyword arguments so they'll do `ct.dispatch(lattice, pos_args=[1, 2], kw_args={x: 3, y: 4}`. Which already looks too convoluted for a simple task like passing in inputs. Any variation of this like `ct.dispatch(lattice, args=(some_pos_args, some_kwargs)` will result in forcing the user to pass their arguments in a certain manner which we want to avoid.

In `ct.dispatch(pipeline)(*args, **kwargs)` they can pass their arguments how they would in a normal function.

covalent/_workflow/lattice.py
kessler-frost: Hmm, why can't we remove it right now? We will be changing the minor patch version indicating that a breaking change has been introduced. Plus I have changed it in the tutorials and how tos so new users won't be affected and old users won't be affected since they don't have this change. Plus, even if I change it in this PR, the only users that are going to see it are those who have installed the latest version. 🤔

CONTRIBUTING.md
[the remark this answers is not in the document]
FyzHsn: Ah great answer, thanks!

FyzHsn (changes requested):
(non-blocking) I'd create an issue to add more in-depth unit tests for the wrappers in the local dispatcher. Also, you might want to put in Deprecation warning like @santoshkumarradha mentioned to give users a chance to adapt and not have all their code break suddenly. However, covalent has only been OS for a week, so it might be fine. 

Syntax needs to be updated in `concepts.rst`.

FyzHsn (dismissed):
This is good to be merged (depending on @santoshkumarradha 's thoughts on the deprecation).

covalent/_workflow/lattice.py
kessler-frost: After some discussion we decided to put the deprecation warning.

santoshkumarradha (dismissed):
💯  !!

santoshkumarradha (approved):
THE BEST ! 🔥 🚀 🤼‍♀️

Issue #60, opened 2022-01-31
Add type hint for result object in lattice


FyzHsn:
Add type hint for `result_object` in `covalent/_workflow/lattice.py` inside the `_server_dispatch` method.

NeelPatel31:
Hey, I want to work on this issue.
Can you explain the problem briefly ?

kessler-frost:
Hi @NeelPatel31! Thank you for sharing your interest in contributing to covalent. Although, it seems someone is already working on this issue, you can still take a look at some of our other good first issues [here](https://github.com/AgnostiqHQ/covalent/labels/good%20first%20issue) and pick one which interests you the most.

PR #68, opened 2022-02-01
Add type hint for result object in lattice.py

Udayan853:
Add type hint for result_object in covalent/_workflow/lattice.py inside the _server_dispatch method.

Issue: https://github.com/AgnostiqHQ/covalent/issues/60


covalent/_workflow/lattice.py
```diff
@@ -343,7 +343,7 @@ def dispatch_sync(self, *args, **kwargs) -> "Result":
 
         return get_result(self.dispatch(**kwargs), self.metadata["results_dir"], wait=True)
 
-    def _server_dispatch(self, result_object) -> str:
+    def _server_dispatch(self, result_object:Result) -> str:
```
FyzHsn, line 346: There should be a space between the parameter and the parameter type. Can you run the tests and ensure that there are no circular import issues? If there are, you might have to use the string representation of result as in line 324.

FyzHsn (changes requested):

covalent/_workflow/lattice.py
Udayan853: Fixed the issues with circular import by using string representation of result, as you had suggested. I also ran the tests to ensure there were no problems regarding the same.

FyzHsn (approved):
LGTM!

FyzHsn (commented):
You just need to update the Changelog and Version files and this is good to be merged. This would count as an increment in the patch version.

Udayan853:
Done!

FyzHsn (approved):

Issue #58, opened 2022-01-31
Add type hints to results class


FyzHsn:
Add type hints to getter methods decorated by `@property` in the Result class in `covalent/_results_manager/results.py`.

PR #70, opened 2022-02-01
fixing result class var type

HaimHorowitzAgnostiq:
Fixes issue #58 


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.22.13] - 2022-02-01
+
+### Fixed
+
+
```
FyzHsn, line 12: Missing description of change in Changelog.

FyzHsn (changes requested):
Need to finish adding the other types.

covalent/_results_manager/result.py
```diff
@@ -109,71 +109,71 @@ def __str__(self):
...
+    def status(self) -> Status:
         """
         Status of current dispatch.
         """
 
         return self._status
 
     @property
-    def result(self):
+    def result(self) -> int:
```
FyzHsn, line 160: I think the result could also be `float`. You might have to do something like, Union[int, float]. Look up Union in the Typing module.

covalent/_results_manager/result.py
```diff
@@ -109,71 +109,71 @@ def __str__(self):
...
+    def results_dir(self) -> str:
         """
         Results directory used to save this result object.
         """
 
         return self._results_dir
 
     @property
-    def lattice(self):
+    def lattice(self) -> Lattice:
```
FyzHsn, line 136: I don't see Lattice being imported anywhere. This is a covalent specific data type. You'd probably need to import it before using it. There might also be issues with circular imports when doing that in which case, you might have to just use `"Lattice"` instead of `Lattice`.

FyzHsn (changes requested):

FyzHsn (changes requested):
@HaimHorowitzAgnostiq You also need to pull in the latest changes from the remote develop branch and fix the merge conflicts. Lastly, you need to update the CHANGELOG and Version files.

covalent/_results_manager/result.py
```diff
@@ -109,71 +109,71 @@ def __str__(self):
...
+    def status(self) -> Status:
         """
         Status of current dispatch.
         """
 
         return self._status
 
     @property
-    def result(self):
+    def result(self) -> Union[int, float]:
```
FyzHsn, line 160: It might also take the types list or perhaps even dict. Could you double check and add that in?

FyzHsn (changes requested):

FyzHsn (approved):

Issue #59, opened 2022-01-31
Add type hint for context manager method


FyzHsn:
Add type hint for `dispatch_info` parameter in `claim` method of `ActiveDispatchInfoManager`

sayandipdutta:
Would like to work on it. Could you assign it to me?

FyzHsn:
> Would like to work on it. Could you assign it to me?

Done!

PR #75, opened 2022-02-02
Add missing type annotations in `covalent\executor\base.py`

### LINKED ISSUE #62: Add type hint for base executor
Some of the fields in the `covalent/executor/base.py` module are missing Type hints. This issue is to ensure that all the fields have type hints.

sayandipdutta:

closes #62 

## Add missing type annotations in `covalent\executor\base.py`
- Removed unused `"typing.Optional"` import
- Replaced `"typing.Optional"` with `"str"`
- Added missing type hints to `BaseExecutor.get_dispatch_context`
- Added missing type hints to `BaseExecutor.write_streams_to_file`


sayandipdutta:
The following could be relevant to #59 
```diff
-  def get_dispatch_context(self, dispatch_info) -> None:
+  def get_dispatch_context(
+      self, 
+      dispatch_info: DispatchInfo
+  ) -> ContextManager[DispatchInfo]:
```

covalent/executor/base.py
```diff
@@ -67,10 +67,10 @@ class BaseExecutor(ABC):
...
     def __init__(
         self,
-        log_stdout: Optional[str] = "",
-        log_stderr: Optional[str] = "",
-        conda_env: Optional[str] = "",
-        cache_dir: Optional[str] = "",
+        log_stdout: str = "",
+        log_stderr: str = "",
+        conda_env: str = "",
+        cache_dir: str = "",
```
jkanem, line 73: Why get rid of the `Optional` annotation?

kessler-frost: @sayandipdutta I know I agreed with it in the comment https://github.com/AgnostiqHQ/covalent/issues/62#issuecomment-1027493779, my bad, I realize now that removing `Optional` here is going to be non-standard. Since, the purpose of `Optional` is to indicate that this parameter already has a default value irrespective of whether it is an empty string or not.

jkanem: Great, I'll approve once this is put back in and we are back up-to-date with develop.

sayandipdutta: I can put back `Optional`, just would request you to check what actual `typing.Optional` docs say about default value type annotations. https://docs.python.org/3/library/typing.html#typing.Optional
> Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the Optional qualifier on its type annotation just because it is optional.

sayandipdutta: And this is not merely due to aesthetics, in my experience, `Optional` qualifiers can create some annoying type checking issues with mypy. For example:
**test.py**
```python
from typing import Optional
def foo(arg: str) -> None:
    pass

def bar(arg: Optional[str] = "") -> None:
    foo(arg)
    return
```
Running `mypy`:
```python
> mypy .\test.py
.\test.py:6: error: Argument 1 to "foo" has incompatible type "Optional[str]"; expected "str"
Found 1 error in 1 file (checked 1 source file)
```
Then you will need to do one of the following:
```python
from typing import cast
def bar(arg: Optional[str] = "") -> None:
    foo(cast(str, arg))
    return
```
**or**
```python
def bar(arg: Optional[str] = "") -> None:
    assert arg is not None
    foo(arg)
    return
```
**or**
```python
def bar(arg: Optional[str] = "") -> None:
    if isinstance(arg, str):
        foo(arg)
    return
```
Although a single `cast` , `assert` or `isinstance` may not be a big deal, if you have to use it repeatedly, especially in loops, it looks unclean and reduces performance.

jkanem: Ok, sounds good. Can you pull in the latest changes from develop, and update the VERSION and CHANGELOG files?

kessler-frost:
@sayandipdutta can you also create an issue referencing these comments, to change the usage of `Optional` in the repo? Because as far as I know that is being used in a different manner than provided in the documentation.

sayandipdutta:
@kessler-frost Sure, I'll add it in half an hour.

jkanem (approved):

jkanem:
Hi, @sayandipdutta have you signed the CLA (https://cla-assistant.io/AgnostiqHQ/covalent)? The CLA check is hanging and may be waiting for you to sign.

sayandipdutta:
@jkanem yeah, noticed that. I did sign it earlier, forced rechecked it again.

Issue #61, opened 2022-01-31
Add type hint in transportable object


FyzHsn:
Add type hint for `data` field in `deserialize` method of `TransportableObject` in `covalent/_workflow/transport.py`.

Udayan853:
Hey FyzHsn, I would like to work on this issue.

FyzHsn:
> Hey FyzHsn, I would like to work on this issue.

Hi @Udayan853 that sounds great! I'm assigning you the issue.

PR #87, opened 2022-02-03
Add type hint to deserialize method in transport.py

Udayan853:
Add type hint to data field in deserialize method of TransportableObject in covalent/_workflow/transport.py

Issue: https://github.com/AgnostiqHQ/covalent/issues/61


covalent/_workflow/transport.py
```diff
@@ -88,7 +88,7 @@ def serialize(self) -> bytes:
         )
 
     @staticmethod
-    def deserialize(data) -> "TransportableObject":
+    def deserialize(data: "TransportableObject") -> "TransportableObject":
```
FyzHsn, line 91: @Udayan853 I'm not a 100% convinced that the data here is of TransportableObject type. Could you try creating an instance of TransportableObject and experiment with what type of inputs the deserialize method can take? Double checking this carefully can't hurt. 

One reason why I think this might not be the data type is bcs cloudpickle is being used to load the data in ln 102. Perhaps, it's more useful to check what are the input types that cloudpickle.loads accepts.

Another point is that perhaps the description of data in args is confusing. Feel free to change it if you think this can be improved.

FyzHsn (changes requested):

covalent/_workflow/transport.py
FyzHsn: Excellent!

Udayan853: @FyzHsn Yeah, data should have been `bytes` type as sayandipdutta stated, it was ignorant of me to have assumed it to be `TransportableObject` blindly.
Would it be better to have the description of args as `base64 encoded cloudpickled function`?

FyzHsn: @Udayan853 I think just “cloud pickled function” as an arg description should be suffice. :)

FyzHsn (approved):
Good job @Udayan853! This is ready to be merged.

PR #88, opened 2022-02-03
requiring todays date in the changelog

wjcunningham7:


wjcunningham7:
on the off-hand chance that the version check does fail, e.g., due to timezones, the action will still print the date it is expecting to see.

scottwn:
Are we sure this isn't going to create weird time zone problems? What if the TZ of the runner is different than the contributor's local TZ?

scottwn (dismissed):

PR #89, opened 2022-02-04
"Change Optional[T] to T, update VERSION to 0.24.2, update CHANGELOG"

### LINKED ISSUE #82: Replace `Optional[T]` with `T` for default args, that expect `T` and not None
In quite a few places, function parameters with *defaults* are annotated with `typing.Optional[T]` (where `T` could be `str`, `int`, ...), I propose changing these to `T` where it is deemed to be the expected type of the variable for the rest of the block, and explicit `None` is not valid.

[`typing.Optional` docs](https://docs.python.org/3/library/typing.html#typing.Optional) says the following about default value type annotations. 
> Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the Optional qualifier on its type annotation just because it is optional.

---

This is not merely due to aesthetics, in my experience, `Optional` qualifiers can create some annoying type checking issues with mypy. For example:
**test.py**
```python
from typing import Optional
def foo(arg: str) -> None:
    pass

def bar(arg: Optional[str] = "") -> None:
    foo(arg)
    return
```
Running `mypy`:
```python
> mypy .\test.py
.\test.py:6: error: Argument 1 to "foo" has incompatible type "Optional[str]"; expected "str"
Found 1 error in 1 file (checked 1 source file)
```
Then you will need to do one of the following:
```python
from typing import cast
def bar(arg: Optional[str] = "") -> None:
    foo(cast(str, arg))
    return
```
**or**
```python
def bar(arg: Optional[str] = "") -> None:
    assert arg is not None
    foo(arg)
    return
```
**or**
```python
def bar(arg: Optional[str] = "") -> None:
    if isinstance(arg, str):
        foo(arg)
    return
```
Although a single `cast` , `assert` or `isinstance` may not be a big deal, if you have to use it repeatedly, especially in loops, it looks unclean and reduces performance.

---
_Originally posted by @sayandipdutta in https://github.com/AgnostiqHQ/covalent/pull/75#discussion_r797989375_

Related: https://github.com/AgnostiqHQ/covalent/pull/75#discussion_r797961991

sayandipdutta:

closes #82 

- Typing: Replace `Optional[bool]` to `bool` as type annotation for `develop` parameter in `covalent_dispatcher.service._graceful_start`
- Typing: Replace `Optional[Any]` to `Any` as type annotation for `new_value` parameter in `covalent._shared_files.config.get_config`


FyzHsn:
@sayandipdutta Let me know whenever this is ready for review :)

sayandipdutta:
@FyzHsn Done merging latest changes, you can check now.

FyzHsn (approved):
LGTM!

Good work @sayandipdutta!

sayandipdutta:
@FyzHsn thanks! Enjoying contributing here 🍻

PR #100, opened 2022-02-09
Fix failing workflow cancellation test

### LINKED ISSUE #93: Test action failed
### Environment

- **Covalent version**: 0.24.5
- **Python version**: 3.8
- **Operating system**: Centos 7

### What is happening?

The test action failed on the develop branch: https://github.com/AgnostiqHQ/covalent/runs/5078639903?check_suite_focus=true

Hard to reproduce.

### What should happen?

Test should pass.

### Any suggestions?

The unit test `test_dispatch_cancellation` has failed because the workflow completes before it can be cancelled.  This will not happen often, as evidenced by the test passing in the other environments.  Suggested remedy is to increase the sleep time to something very large, to ensure it absolutely cannot complete before cancellation.

cjao:
Increase sleep duration to 600 seconds to prevent workflow
from completing before cancellation.

Closes #93.


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.24.9] - 2022-02-09
+
+- Fixed failing workflow cancellation test.
```
FyzHsn, line 10: The format should be:

```
### Fixed

- Failing workflow cancellation test in [test_location].

```

FyzHsn (changes requested):

CHANGELOG.md
cjao: Thanks for the correction, fixed.

FyzHsn (approved):

FyzHsn (dismissed):
(non-blocking) You need to update the Version number and the Changelog. Then it's good to go!

FyzHsn (approved):

PR #101, opened 2022-02-10
Errors in executor init

jkanem:


wjcunningham7:
@jkanem can you apply the changes in this diff in this PR?  

```
diff --git a/covalent/_shared_files/config.py b/covalent/_shared_files/config.py
index 81f5c1e..5056bf2 100644
--- a/covalent/_shared_files/config.py
+++ b/covalent/_shared_files/config.py
@@ -24,7 +24,7 @@ import shutil
 from functools import reduce
 from operator import getitem
 from pathlib import Path
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Optional, Union
 
 import toml
 
@@ -74,27 +74,35 @@ class _ConfigManager:
         # self.config_data may be modified later, and we don't want it to affect _DEFAULT_CONFIG
         self.config_data = copy.deepcopy(_DEFAULT_CONFIG)
 
-    def update_config(self) -> None:
+    def update_config(self, new_entries: Optional[Dict] = None, defaults: bool = False) -> None:
         """
         Update the exising configuration dictionary with the configuration stored in file.
 
         Args:
-            None
+            new_entries: Dictionary of new entries added or updated in the config
+            defaults: Defaults do not overwrite existing entries
 
         Returns:
             None
         """
 
-        def update_nested_dict(old_dict, new_dict):
+        def update_nested_dict(old_dict, new_dict, defaults):
             for key, value in new_dict.items():
                 if isinstance(value, dict) and key in old_dict and isinstance(old_dict[key], dict):
-                    update_nested_dict(old_dict[key], value)
+                    update_nested_dict(old_dict[key], value, defaults)
                 else:
-                    old_dict[key] = value
+                    if defaults:    # Values provided are defaults
+                        old_dict.setdefault(key, value)
+                    else:           # Values provided should overwrite
+                        old_dict[key] = value
 
         if os.path.exists(self.config_file):
             file_config = toml.load(self.config_file)
             update_nested_dict(self.config_data, file_config)
+            if new_entries:
+                update_nested_dict(self.config_data, new_entries, defaults)
+
+        self.write_config()
 
     def read_config(self) -> None:
         """
@@ -163,7 +171,14 @@ class _ConfigManager:
         """
 
         keys = key.split(".")
-        reduce(getitem, keys[:-1], self.config_data)[keys[-1]] = value
+        try:
+            reduce(getitem, keys[:-1], self.config_data)[keys[-1]] = value
+        except KeyError:
+            data = self.config_data
+            for key in keys:
+                data[key] = {}
+                data = data[key]
+            data[keys[-1]] = value
 
 
 _config_manager = _ConfigManager()
```

The change to `update_config` allows me to do this in a separate module:

```
my_module_config_defaults = {
    "key1": "value1"
}
from covalent._shared_files.config import _config_manager as cm
cm.update_config(my_module_config_defaults, True)
```

which will add these defaults to the config only if they were not already provided in the config file.

The change to `set` allows the user to do
```
import covalent as ct
ct.set_config("my.new.nested.key", "new_value")
```
for a nested key which doesn't already exist, without throwing a `KeyError` (the current behavior).

jkanem:
> @jkanem can you apply the changes in this diff in this PR?
Done

FyzHsn:
@jkanem @wjcunningham7 I've added the minimum amount of tests required to ensure that diffs added in this branch are covered. @jkanem lemme know when this is ready for another review.

jkanem:
@FyzHsn , I believe this is ready for review, thanks.

FyzHsn (approved):

PR #108, opened 2022-02-11
Containerization of covalent by merging flask servers 

AlejandroEsquivel:


VERSION
```diff
@@ -1 +1 @@
-0.24.16
+0.25.16
```
scottwn, line 1: should be `0.25.0` or `0.24.17`

USING_DOCKER.md
```diff
@@ -0,0 +1,33 @@
+# Covalent Docker Guide
```
scottwn, line 1: This guide should be linked to from the README, or included directly in the README since it's nice and short. Same for readthedocs.

scottwn (changes requested):

CONTRIBUTING.md
```diff
@@ -36,6 +36,29 @@ Finally, run the tests to verify your installation:
...
+
+### Running Container
+
+The following would be used to run the covalent docker container:
+
+```shell
+docker run -d -p 48008:8080 covalent:latest
+```
+
+You should be able to see the dashboard at [http://0.0.0.0:47008](http://0.0.0.0:47008).
```
scottwn, line 55: Is this right? We're forwarding to `48008` in `docker run` but the UI is still on `47008`?

scottwn (changes requested):

CONTRIBUTING.md
AlejandroEsquivel: I think the port would be 47008 but I agree it's a tad confusing, I can update it to default to 47007 to be more consistent given that the UI used to run at that port. I think the reason it starts at 47008 is that it is the default value for the `--port` arg for the `covalent start` command at the moment.

doc/source/getting_started/index.rst
```diff
@@ -35,18 +35,26 @@ Users can also install Covalent as a package in a Conda environment:
 Docker Install
 --------------
 
-Covalent is also provided as a Docker image. This image can be used as a developer environment or simply to run the Covalent dispatch server in a container:
+Covalent is also provided as a Docker image. You can pull the covalent server image with the following command: 
 
 .. code:: bash
 
    docker pull public.ecr.aws/i9y7r1d8/covalent
```
scottwn, line 42:
```suggestion
   docker pull public.ecr.aws/covalent/covalent
```

Our custom ECR public alias was approved so we should stop using this programmatic default

scottwn (changes requested):

doc/source/getting_started/index.rst
AlejandroEsquivel: Gotcha 👍

AlejandroEsquivel:
@scottwn @wjcunningham7 Currently the server can start with the cli or Docker, but the remaining work would be updating all the docs and references in the cli to the individual components (UI & Dispatch server) to now name one unified component "Covalent Server" (As well as updating the config file). Besides that the only thing I see is that the CLI doesn't resolve the correct port of the server when doing `covalent status`. 

If it is considered out of scope for this issue then this issue can be closed, and a new issue opened with these requirements.

scottwn:
I think breaking `covalent status` is a regression, so we need to fix that before we merge this, and update the docs as well

cjao:
`covalent status` doesn't work because the server process listens to multiple ports:
* The web server's port, which we actually want
* The ports used by Dask's LocalCluster.

This wasn't a problem before because we were using `gunicorn`, which runs LocalCluster in a worker process and not the master process associated with the pidfile. Now that we are running `covalent_ui/app.py` directly, we need some other way to track the relevant port. Of course going back to `gunicorn` would also fix the regression, but presumably there were reasons for ditching `gunicorn`.

cjao:
Do we support users editing the config file by hand? If so, users should be cautioned that the dispatcher and user interface ports should be the same.

cjao:
For some reason all tests pass except the very first run of `test_arithmetic_1`.

cjao:
Ready for review.

scottwn:
FWIW I was able to reproduce the [failed test](https://github.com/AgnostiqHQ/covalent/runs/5324421145?check_suite_focus=true) on [ee88a85](https://github.com/AgnostiqHQ/covalent/pull/108/commits/ee88a85840f868b66445bc1d0bb6e6456ffe0ceb)
```
$ docker build -t covalent-test .   
$ docker run --rm -it --entrypoint bash covalent-test    
# pip install -r tests/requirements.txt  
# covalent start -d
# pytest -v
```

covalent_dispatcher/_cli/service.py
```diff
@@ -231,9 +231,18 @@ def start(ctx, port: int, develop: bool) -> None:
...
-        time.sleep(3)
+
+    # Wait until the server actually starts listening on the port
+    server_listening = False
+    sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    while not server_listening:
+        try:
+            sock.bind(("0.0.0.0", port))
+            sock.close()
+        except:
```
kessler-frost, line 242: Small comment to catch the specific exception that will be thrown if `try` block fails rather than any exception. Even in the case where it is unknown which kind of exception will be raised we should use `Exception` and not an empty `except` clause.

scottwn (changes requested):
Documentation of port numbers is inconsistent, see
https://github.com/AgnostiqHQ/covalent/pull/108/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R61
https://github.com/AgnostiqHQ/covalent/pull/108/files#diff-9e3f76f5b7c0a6e9a6f51b465ac261609619c98cadb759409660450f969418ffR60
https://github.com/AgnostiqHQ/covalent/pull/108/files#diff-63e953e61bd0f9a6dca5cda1219bffc24d0c1b614d23583e3360690f7c9f7ae1R128
https://github.com/AgnostiqHQ/covalent/pull/108/files#diff-63e953e61bd0f9a6dca5cda1219bffc24d0c1b614d23583e3360690f7c9f7ae1R159

I think we need to double check which port is really being used and that the docs reflect that

cjao:
I think I've changed all references to port numbers other than 48008 (except for in the older changelog entries).

README.md
```diff
@@ -58,7 +58,7 @@ Begin by starting the Covalent servers:
 covalent start
 ```
 
-Navigate to the user interface at `http://localhost:47007` to monitor workflow execution progress.
+Navigate to the user interface at `http://localhost:47008` to monitor workflow execution progress.
```
kessler-frost, line 61: I think you missed this one @cjao 😅 .

cjao: Thanks for the catch.

USING_DOCKER.md
scottwn: @AlejandroEsquivel was this linked? I didn't see a link in the README but maybe I missed it

scottwn (approved):

USING_DOCKER.md
AlejandroEsquivel: No I don't think it's linked, I actually intended to delete this

scottwn: ah OK I think deleting it is fine. should we include some Docker information in the README or do you think having it in getting started is enough?

AlejandroEsquivel:
Hmm, it might be a good idea to include a snippet in the README install section, I suppose it’s quite key for those that cannot run with the pip install alone (Windows users for example). I can do that 👍

scottwn (approved):

PR #110, opened 2022-02-11
Fixing output resolution in `post_process`

### LINKED ISSUE #109: `post_process` fails if there is a conflict in node names
An example of this is:

```python
@ct.electron
def a(x):
    return x,x**2

@ct.lattice
def p(x):
    result,b=a(x=x)
    for _ in range(1):
        result,b=a(x=result)
    return b,result
```

Here, since `post_process` takes in a list of node outputs which also contain non executable electron nodes such as of `:parameter:` type, it fails because it will look for the function name of `a` and find its occurrence in `:parameter:` and then use the output of that for post processing.

This can be avoided if we use an `OrderedDictionary` to store the nodes and their outputs and then pass it to `post_process`.

kessler-frost:
Resolves #109.


santoshkumarradha:
Sweet ! @kessler-frost  do we need a test for this bug ?

kessler-frost:
I added a test.

santoshkumarradha (approved):

Issue #112, opened 2022-02-14
MNIST classifier tutorial


FyzHsn:
Write a beginner tutorial.

PR #118, opened 2022-02-15
MNIST classifier tutorial

FyzHsn:
Fixes issue #112 

- [ ] Plan the MNIST classifier without using Covalent.
- [ ] Plan the MNIST classifier without using Covalent.
- [ ] Write wireframe of section w/o Covalent
- [ ] Write wireframe of section w Covalent
- [ ] Add & troubleshoot code w/o Covalent
- [ ] Add & troubleshoot code w Covalent
- [ ] Tutorial writeup
- [ ] Ask for review from software team.
- [ ] Ask for review from research team.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [UNRELEASED]
+
+## Added
+
+-
+
 ## [0.24.17] - 2022-02-15
 
 ### Added


FyzHsn:
@HaimHorowitzAgnostiq This is the MNIST classifier tutorial. If you go to files changed and click on view files as shown below, you should be able to see the rendered notebook.
Screen Shot 2022-02-23 at 5 48 22 PM

FyzHsn:
@kessler-frost This is at a stage for you to give a review.

kessler-frost (commented):
I have added some comments in the notebook. They can be found with the keyword `# TODO` when you load it locally. Other than those, I think this is pretty good. Also, maybe a slight cleanup of the last few lines will be needed in the notebook once reviewed by others.

HaimHorowitzAgnostiq (commented):
1. In the "Run MNIST classifier workflow" section, unless it's necessary for some reason, I would remove everything between Epoch 1 and the actual final result.
2. There is a discrepancy between the screenshot and the printed results above it (?).
3. In the first bullet point of the conclusion, you might want to address what's in the square brackets.
4. In the second bullet point, "what we can be done" should be "what can be done".
5. It might be nice to print also the time taken to train and test the model using Covalent.

FyzHsn:
> I have added some comments in the notebook. They can be found with the keyword `# TODO` when you load it locally. Other than those, I think this is pretty good. Also, maybe a slight cleanup of the last few lines will be needed in the notebook once reviewed by others.

Thanks, the comments were very helpful.

FyzHsn:
> 1. In the "Run MNIST classifier workflow" section, unless it's necessary for some reason, I would remove everything between Epoch 1 and the actual final result.

I'll keep this just bcs it's standard to show the training progress over the epochs.

> 2. There is a discrepancy between the screenshot and the printed results above it (?).

I'll keep the output and remove the screenshot.
> 3. In the first bullet point of the conclusion, you might want to address what's in the square brackets.
Done
> 4. In the second bullet point, "what we can be done" should be "what can be done". 
Good catch
> 5. It might be nice to print also the time taken to train and test the model using Covalent.
Great idea, but I'll skip due to time constraints

Thanks for the review!

wjcunningham7 (changes requested):
- I had to also install `ipywidgets` and use the line `!jupyter nbextension enable --py widgetsnbextension` in the notebook.
- The workflow graph in the UI screenshot needs updating
- The user interface should consistently be referred to as the "Covalent UI", not browser or viewer.
- Emphasize the difference in execution time in the conclusion

FyzHsn:
> * I had to also install `ipywidgets` and use the line `!jupyter nbextension enable --py widgetsnbextension` in the notebook.

Great, I’ll these as notes.

> * The workflow graph in the UI screenshot needs updating

Has there been a UI update?

> * The user interface should consistently be referred to as the "Covalent UI", not browser or viewer.
> * Emphasize the difference in execution time in the conclusion

Should we emphasize the difference in execution time here? Except for the data loading part the rest has to happen sequentially and we can’t really make it faster than the normal workflow.

wjcunningham7 (approved):

PR #131, opened 2022-02-21
Add leptons

wjcunningham7:

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -20,6 +20,6 @@
 
 [flake8]
 max-line-length = 99
-max-complexity = 18
+max-complexity = 21
 select = B,C,E,F,W,T4,B9
 ignore = E501, E722, W503, F401, F403, F811, F841, E203
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,5 @@
+all:
+	gcc -shared -fPIC -o libtest.so test.c
+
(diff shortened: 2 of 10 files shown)


santoshkumarradha:
🔥

wjcunningham7:
next steps:
- refactor some of lepton.py so it's more modular and testable
- write integration tests with a set of example functions and expected input-output pairs (may end up in subsequent issue)

FyzHsn:
One of the cells gives me an error:
Screen Shot 2022-03-01 at 11 52 10 PM

When I remove the -D tag, the cell runs fine, but the workflow dispatching returns a 500 server error.

wjcunningham7:
> One of the cells gives me an error: Screen Shot 2022-03-01 at 11 52 10 PM
> 
> When I remove the -D tag, the cell runs fine, but the workflow dispatching returns a 500 server error.

I'm testing on Linux so no guarantee `nm` behaves the same on mac.  Also for the example to work you need to make sure the library is on your `LD_LIBRARY_PATH`, else you can manually specify the absolute path in the notebook.

wjcunningham7:
Also make sure it's compiling with actual gcc, not clang or some Mac alternative.

FyzHsn (approved):

Issue #77, opened 2022-02-02
Improve automation for versioning and changelog


scottwn:
@scottwn commented on [Wed Nov 24 2021](https://github.com/AgnostiqHQ/covalent-staging/issues/119)

In GitLab by @scottwn on Nov 15, 2021, 14:11

## Overview
It might be helpful to use tagged commit messages to automatically generate changelog and semantic version.

## Technical Details
Currently GitLab CI has limitations that cause issues with this. It is not possible for the squash commit to automatically include all the commit messages from the branch. See https://gitlab.com/gitlab-org/gitlab/-/issues/26303

If the squash commit issue can be resolved, then we can use a tool like https://github.com/relekang/python-semantic-release

## Expected Results
Versioning and changelog will be controlled automatically by commit messages rather than by hand.

scottwn:
This will change the way changelog entries are processed.

Instead of heading the changelog entry with the version and date, it will be headed with `[UNRELEASED]` and the automation will handle the version and date.

PR #132, opened 2022-02-22
add changelog reminder

scottwn:

TODO: modify the version action per #77


FyzHsn (commented):
@scottwn Am I correct to assume that this reminder is invoked if the user pushes the local branch to remote without changing the CHANGELOG?

FyzHsn (dismissed):

VERSION
```diff
@@ -1 +1 @@
-0.25.1
+0.26.0
```
kessler-frost, line 1: Just curious for my understanding, why did you update the minor version number instead of only the patch version?

scottwn: because it's new functionality rather than a fix

FyzHsn (approved):

wjcunningham7 (approved):

PR #134, opened 2022-02-23
adding named arguments inside kwargs

wjcunningham7:


FyzHsn (commented):
@wjcunningham7 Looks good. Could you update one or two of the functional tests so that it uses both args and kwargs?

PR #138, opened 2022-02-25
fix ui inputs (derived from graph) and executor rename

valkostadinov:
- Missing UI input data after backend change - needed to be derived from graph for electrons, lattice inputs fixed on server-side, combining name and positional args
- Broken UI graph due to variable->edge_name renaming
- Missing UI executor data after server-side renaming


kessler-frost (changes requested):
Looks pretty good, one thing though, in here:

Screen Shot 2022-02-28 at 2 16 52 AM

The parameters don't appear to be connected to the `task` node. That isn't the expected behavior right?

valkostadinov:
> Looks pretty good, one thing though, in here:
> The parameters don't appear to be connected to the `task` node. That isn't the expected behavior right?

This bug has proven to be a tough one to fix. It seems to show up only when there's a single electron node. The moment I add a new one, it goes away. Also, if I change direction, the edges show up correctly. I've tracked it down to a warning message in the react-flow library and it seems like a library bug related to the state but it's hard to confirm. I'm currently unable to find a workaround.

Let's push this PR as it fixes other problems and file a separate bug report just for the missing edges.

kessler-frost (approved):

PR #149, opened 2022-03-03
Skeleton code for refactor with OpenAPI Spec

kessler-frost:

This PR will introduce the different microservices for the refactoring of dispatcher component. Also includes the openapi spec for each of the service.


refactor/data/openapi.yaml
```diff
@@ -0,0 +1,344 @@
...
+  /api/v0/fs/upload:
+    post:
+      tags:
+        - Data
+      summary: Upload File
+      description: Upload a file
+      operationId: upload_file_api_v0_fs_upload_post
+      requestBody:
+        content:
+          multipart/form-data:
```
scottwn, line 109: Is this the right `POST` data type for the server to pass a `FileStorage` object to the backend? I think this is the type we want to use to wrap files https://werkzeug.palletsprojects.com/en/2.0.x/datastructures/#werkzeug.datastructures.FileStorage

refactor/dispatcher/openapi.yaml
```diff
@@ -0,0 +1,189 @@
+openapi: 3.0.2
+info:
+  title: Covalent Dispatcher Service API
+  version: 0.1.0
+paths:
+  /api/v0/workflow/{dispatch_id}/dispatch:
```
scottwn, line 6:
```suggestion
  /api/v0/workflow/{dispatch_id}:
```

refactor/dispatcher/openapi.yaml
```diff
@@ -0,0 +1,189 @@
...
+              schema:
+                $ref: '#/components/schemas/DispatchWorkflowResponse'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /api/v0/workflow/{dispatch_id}/cancel:
+    post:
```
scottwn, line 34:
```suggestion
    delete:
```

refactor/dispatcher/openapi.yaml
```diff
@@ -0,0 +1,189 @@
...
+              schema:
+                $ref: '#/components/schemas/CancelWorkflowResponse'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /api/v0/workflow/{dispatch_id}/update:
+    post:
```
scottwn, line 61:
```suggestion
    put:
```

refactor/runner/openapi.yaml
```diff
@@ -0,0 +1,168 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RunTaskResponse'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /api/v0/workflow/{dispatch_id}/task/{task_id}/status:
```
scottwn, line 42:
```suggestion
  /api/v0/workflow/{dispatch_id}/task/{task_id}:
```

refactor/runner/openapi.yaml
```diff
@@ -0,0 +1,168 @@
...
+              schema:
+                $ref: '#/components/schemas/TaskStatus'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /api/v0/workflow/{dispatch_id}/task/{task_id}/cancel:
+    delete:
```
scottwn, line 76:
```suggestion
    delete:
```

scottwn (changes requested):

AlejandroEsquivel:
Totally concur with these changes @scottwn , will update the specs accordingly

refactor/data/openapi.yaml
AlejandroEsquivel: I think the intended use case of this endpoint is something like:

```
import requests

files = {'file': ('foo.txt', open('./foo.txt', 'rb'))}
response = requests.post('http://127.0.0.1:8000/api/v0/fs/upload', files=files)
print(response)
print(response.json())
```
the fast API server
```
from fastapi import FastAPI, File, UploadFile
import uvicorn

app = FastAPI()

@app.post('/file')
def _file_upload(my_file: UploadFile = File(...)):
    print(my_file)

if __name__ == "__main__":
    uvicorn.run("main:app", host="0.0.0.0", port=8000, log_level="debug")
```

source: https://stackoverflow.com/questions/65102579/send-and-receive-file-using-python-fastapi-and-requests

scottwn: I did a little googling on this, and I think you're right that `multipart/form-data` is the correct content type. Although I'm not sure about the `_file_upload` implementation above, I think there might be some issues with that, but for the skeleton I think it's OK.

refactor/dispatcher/openapi.yaml
```diff
@@ -0,0 +1,189 @@
...
+              schema:
+                $ref: '#/components/schemas/DispatchWorkflowResponse'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /api/v0/workflow/{dispatch_id}/cancel:
+    delete:
```
scottwn, line 34:
```suggestion
    delete:
```

refactor/dispatcher/openapi.yaml
```diff
@@ -0,0 +1,189 @@
...
+              schema:
+                $ref: '#/components/schemas/CancelWorkflowResponse'
+        '422':
+          description: Validation Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+  /api/v0/workflow/{dispatch_id}/update:
+    put:
```
scottwn, line 61:
```suggestion
    put:
```

refactor/queuer/openapi.yaml
```diff
@@ -0,0 +1,80 @@
...
+      tags:
+        - Submit
+      summary: Submit Workflow
+      description: Submit a workflow
+      operationId: submit_workflow_api_v0_submit_dispatch_post
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ResultPickle'
```
scottwn, line 17: Is this the right schema to submit to the queuer? We're submitting a workflow, not a result... I think?

refactor/runner/openapi.yaml
```diff
@@ -0,0 +1,168 @@
...
+        - required: true
+          schema:
+            title: Dispatch Id
+            type: string
+          name: dispatch_id
+          in: path
+      requestBody:
+        content:
+          application/json:
+            schema:
```
scottwn, line 23: The doc should include a more robust example, or multiple examples (something more than just `[0]`), and some description of how to do this properly.

It also might be more stylistically consistent to make it
```
{
    'task_ids': [int]
}
```
instead of just the array... I feel like people expect JSON to always be a dict at the top level, even if it's technically OK to have an array at the top level

scottwn (changes requested):

refactor/data/app/api/api_v0/endpoints/fs.py
```diff
@@ -0,0 +1,63 @@
...
+from app.schemas.common import HTTPExceptionSchema
+from app.schemas.fs import UploadResponse
+from fastapi import APIRouter, File, HTTPException
+from fastapi.responses import FileResponse
+
+router = APIRouter()
+
+
+@router.post("/upload", status_code=200, response_model=UploadResponse)
+def upload_file(*, file: bytes = File(...)) -> Any:
```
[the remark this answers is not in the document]
AlejandroEsquivel: According to the FastAPI docs I believe it should be set to `bytes` if we want to have the file contents in memory, however I am not totally familiar with the FileStorage object you mentioned. The docs also mention the `UploadFile` type annotation which uses a `Spoofed File`, which is described as "A file stored in memory up to a maximum size limit, and after passing this limit it will be stored in disk." - https://fastapi.tiangolo.com/tutorial/request-files/#file-parameters-with-uploadfile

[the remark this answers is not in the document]
scottwn: ok cool, if that's what the FastAPI docs say then let's do it

refactor/queuer/openapi.yaml
scottwn: ok, can that be included in the description of the API method? otherwise I think this is sort of confusing
Screen Shot 2022-03-07 at 4 33 49 PM

refactor/dispatcher/openapi.yaml
kessler-frost: Implemented.

refactor/dispatcher/openapi.yaml
kessler-frost: Implemented.

refactor/queuer/openapi.yaml
kessler-frost: Implemented.

refactor/runner/openapi.yaml
scottwn: Looks like the type also changed here and now it's a string array? Is there a reason that tasks are strings instead of ints?

refactor/runner/openapi.yaml
```diff
@@ -0,0 +1,176 @@
...
+        - required: true
+          schema:
+            title: Dispatch Id
+            type: string
+          name: dispatch_id
+          in: path
+        - required: true
+          schema:
+            title: Task Id
+            type: integer
```
scottwn, line 56: We should be consistent with the type of task IDs, it's changed to `string` in the `POST` call.

refactor/runner/openapi.yaml
```diff
@@ -0,0 +1,176 @@
...
```
scottwn, line 88: same as above... need to be consistent with task ID type

refactor/ui_backend/openapi.yaml
```diff
@@ -0,0 +1,77 @@
...
+        - required: true
+          schema:
+            title: Dispatch Id
+            type: string
+          name: dispatch_id
+          in: path
+        - required: true
+          schema:
+            title: Task Id
+            type: integer
```
scottwn, line 23: task ID should be the same type it is in the runner API https://github.com/AgnostiqHQ/covalent/pull/149/files#diff-db80e5d84802cda9ffede916caa8a32c55476dd117106ff0f99189a7ea19d311R23

scottwn (changes requested):
Looks good except there seems to be some inconsistency with task ID types. They are ints everywhere except `POST` `/api/v0/workflow/{dispatch_id}/tasks`

The `POST` should also include some description or a working example for how to provide the array of tasks.

refactor/runner/openapi.yaml
scottwn: ok cool that makes sense... and how is the id, function, input, etc represented in the string?

kessler-frost: All of this will be 1 pickle object which is bytes, it says "string" but its actually a bytes type.

refactor/runner/openapi.yaml
```diff
@@ -0,0 +1,176 @@
...
+      required:
+        - tasks
+      type: object
+      properties:
+        tasks:
+          title: Tasks
+          type: array
+          items:
+            type: string
+            format: binary
```
kessler-frost, line 162: Here, its mentioned.

scottwn (approved):

scottwn (approved):

PR #164, opened 2022-03-09
Config improvements

jkanem:


jkanem:
Improvements of note are:
- If an executor plugin has the global dict `_EXECUTOR_PLUGIN_DEFAULTS` defined, these defaults are loaded into the in-memory configuration data and written to the configuration file. But only for parameters which were not already defined.
- Usage of `_shared_files.config.update_config` is clearer. With no arguments, the in-memory config is updated with values from the configuration file (no change there). With an input dictionary, the in-memory config and the configuration file are updated with values from the input dictionary. Unless `override_existing == False`, in which case, values from the input dictionary are only used if the parameter in question was either not defined, or has a null value in the current config.

wjcunningham7 (approved):

PR #168, opened 2022-03-10
Allow empty fetchresult

cjao:
- [ x ] I have added the tests to cover my changes.
- [ x ] I have updated the documentation, VERSION, and CHANGELOG accordingly.
- [ x ] I have read the CONTRIBUTING document.

`covalent_ui.app.fetch_result` previously assumed that the requested result exists in the dispatch db. This is not always the case, as the result might not yet have hit the db when the UI is notified of the result. 

With this PR, `fetch_result` checks if the db actually contains the desired result and returns an empty response if not. Previously an unhandled IndexError would be raised by the expression `res[0][1]` if `res` is an empty list.


valkostadinov:
Looks good. Let's return a 400 status code though.

valkostadinov (approved):
That's better. Looking good.

PR #180, opened 2022-03-10
OS Refactor - UI Backend

AlejandroEsquivel:


refactor/ui_backend/app/api/api_v0/endpoints/ui.py
```diff
@@ -17,18 +17,40 @@
...
+    global dispatch_set
+    if (dispatch_id, task_id) not in dispatch_set:
+        dispatch_set.add((dispatch_id,task_id))
+        async with limiter:
+            await notify_frontend(app, dispatch_id, task_id)
+        
 
 @router.put("/workflow/{dispatch_id}/task/{task_id}", status_code=200, response_model=UpdateUIResponse)
-def update_ui(*, dispatch_id: str, task_id: int) -> UpdateUIResponse:
+async def update_ui(*, dispatch_id: str, task_id: int, request: Request, background_tasks: BackgroundTasks) -> UpdateUIResponse:
```
scottwn, line 50: are `request` and `background_tasks` meant to be query params? if so the swagger spec will need to be updated https://app.swaggerhub.com/apis/agnostiq/Covalent_UI_Service_API/0.1.0#/UI/update_ui_api_v0_ui_workflow__dispatch_id__task__task_id__put

scottwn (changes requested):

refactor/ui_backend/app/api/api_v0/endpoints/ui.py
AlejandroEsquivel: @scottwn `request` and `background_tasks` are fastapi dependencies that are injected into all route handlers. `request` is able to get the application context with `request.app` (to avoid circular dependencies), and `background_tasks` is a way to run async coroutines in the background. Hence we do not need to update the api spec as a result.

scottwn (approved):

PR #197, opened 2022-03-15
185 object store

### LINKED ISSUE #185: Object store service
### What should we add?

Implement a service for managing result objects and other files. The service will consist of 2 API methods, `upload` and `download`.

https://app.swaggerhub.com/apis/agnostiq/Covalent_Data_Service_API/0.1.0#/Data

Ideally the service should be able to integrate with a local object storage service or a managed cloud service like S3. For now it is enough to get the local object storage service working. We will need to do some research to find a service that works best.

Objects in the service will need unique IDs that can be stored in the results DB and used by the data service to retrieve results.

We considered storing results directly in the DB. This design was abandoned because results objects can be very large which would impact DB performance and maintainability.

We also considered uploading and downloading files to and from a file system rather than an object store. We agreed on the object store because it should perform better, is easier to implement, and because there is not an immediate need for the service to be able to manage a folder structure directly.

cjao:

Closes #185 .


refactor/data/app/api/api_v0/endpoints/fs.py
```diff
@@ -19,31 +19,51 @@
...
+from pathlib import Path
 from typing import Any
 
 from app.schemas.common import HTTPExceptionSchema
 from app.schemas.fs import UploadResponse
-from fastapi import APIRouter, File, HTTPException
-from fastapi.responses import FileResponse
+from fastapi import APIRouter, File, HTTPException, UploadFile
+from fastapi.responses import StreamingResponse
+from minio import Minio
```
scottwn, line 29: need to add minio in requirements

scottwn (changes requested):

cjao:
A couple of issues to be addressed before this is merged:

- Config management: how should the user configure which StorageBackend to use (Local, Minio) and set backend-specific settings (such as AWS credentials)?
- Better logging

scottwn (approved):

PR #206, opened 2022-03-17
Lepton fix (fixes #205)

### LINKED ISSUE #205: Leptons use path relative to the runtime PWD
### Environment

- **Covalent version**:  0.32.3
- **Python version**:  3.8.12
- **Operating system**:  Ubuntu 21.10

### What is happening?

When trying to use an external library in the lepton wrapper, the path to the module is relative to the present-working-directory (PWD) of the code at runtime. This happens to be the covalent_ui directory. As a result, the library will not be found, unless the user supplies a library path accommodating this starting-point.

Try to use a library in a lepton and don't supply a relative path, with respect to the covalent_ui directory.

### What should happen?

If a full path is supplied, we should not get a `ModuleNotFoundError` error

### Any suggestions?

_No response_

jkanem:
supplying a relative path with respect to the covalent_ui
directory.


jkanem:
The path to the external library should be absolute. Examples, tutorials and how-tos should be updated to reflect this.

FyzHsn (approved):
I'm approving this on the basis that the lepton how-to guide runs fine on Beehive. However, I must mention that the lepton how-to ran even on the develop branch without these changes.

wjcunningham7 (dismissed):

FyzHsn (approved):

PR #215, opened 2022-03-24
Results manager service

scottwn:

I think this is basically done, but I have some outstanding issues that I need help with.

* ~When I try to run this service and the data service at the same time I get `Address already in use` error. How do I fix that?~
* There seem to be issues with the data service download endpoint which will cause problems with this service.
* I'd like some guidance on the best way to update a node. The `_update_node` function seems like it's not meant to be called externally.


scottwn:
Follow up documented in #221 #222

scottwn:
RTD failing due to https://github.com/jupyter/nbconvert/issues/1736

refactor/results/app/api/api_v0/endpoints/workflow.py
```diff
@@ -0,0 +1,219 @@
...
+from fastapi import APIRouter, HTTPException, Request, UploadFile
+from fastapi.responses import FileResponse, StreamingResponse
+
+logging.config.fileConfig("logging.conf", disable_existing_loggers=False)
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter()
+
+fs_server_address = os.environ.get("FS_SERVER_ADDRESS")
```
wjcunningham7, line 46: this should probably call `get_config` and reference the config file.  You could add something like this to this file:

```
_WORKFLOW_DEFAULTS = {
    "service": {
        "fs": {
            "address": os.environ.get("COVALENT_FS_SERVER_ADDRESS") or "localhost",
            "port": int(os.environ.get("COVALENT_FS_PORT")) or 8000,
        }
    }
}

update_config(_WORKFLOW_DEFAULTS, override_existing=False)
```

which would add the following entry to the config, if it does not exist:
```
[service.fs]
address = "localhost"
port = 8000
```

refactor/results/app/api/api_v0/endpoints/workflow.py
```diff
@@ -0,0 +1,219 @@
...
+    raise HTTPException(status_code=500, detail="Error adding record to database.")
+
+
+def _handle_error_response(status_code: int, response: dict):
+    if status_code >= 400:
+        raise HTTPException(status_code=status_code, detail=response["detail"])
+
+
+def _db(sql: str, key: str = None) -> Optional[Tuple[Union[bool, str]]]:
+    results_db = os.environ.get("RESULTS_DB")
```
wjcunningham7, line 116: same about config here

refactor/results/requirements.txt
```diff
@@ -0,0 +1,66 @@
+alembic==1.7.6
```
wjcunningham7, line 1: can this list be trimmed down at all?  we should at least separate out dev requirements (which right now go in tests/requirements.txt) from core user reqs

setup.py
```diff
@@ -83,19 +83,43 @@ class BuildUI(Command):
...
         if self.clean:
             import shutil
 
             shutil.rmtree("covalent_ui/webapp/build", ignore_errors=True)
-
-        else:
+        elif self._node_lts(node_versions):
             import subprocess
 
             subprocess.run(
```
wjcunningham7, line 125: for each of these subprocess.run commands, we should add this:
```
if proc.returncode != 0:
    raise Exception(proc.stderr.decode("utf-8").strip())
```
where `proc` is returned from `subprocess.run`.  Otherwise we might miss errors in the subprocess.

wjcunningham7 (changes requested):

refactor/results/app/api/api_v0/endpoints/workflow.py
AlejandroEsquivel: I think after the supervisord work is merged into `develop` we can simply use environment variables as the `supervisord.conf` (supervisord config file) will manage the environment variables for all the services

wjcunningham7: Great, disregard my comment here then.

refactor/results/app/api/api_v0/endpoints/workflow.py
scottwn: followup documented in #222

refactor/results/app/api/api_v0/endpoints/workflow.py
cjao: So is the consensus here that env variables are the way to go?

kessler-frost (dismissed):

PR #269, opened 2022-03-31
Replace TemporaryFile with io.BytesIO in `update_result`

kessler-frost:


scottwn (changes requested):
I don't think the data service will accept the `BytesIO`. We need to decide where the conversion from bytes to file happens.
There are related changes in 
https://github.com/AgnostiqHQ/covalent/pull/278
https://github.com/AgnostiqHQ/covalent/pull/275
https://github.com/AgnostiqHQ/covalent/pull/272
https://github.com/AgnostiqHQ/covalent/pull/271

kessler-frost:
I just checked the `refactor-intermediate` branch, and it seems there in `refactor/upload_pkl_file.py` I'm using,

```python
response = requests.post(
    url=url_endpoint, files={"result_pkl_file": BytesIO(pickle.dumps(result_object))}
)
```

to send the file to the result service, that endpoint expects `UploadFile` type object (can be seen in `refactor/results/app/api/api_v0/endpoints/workflow.py`) and that seems to work.

And as we can see in `refactor/data/app/api/api_v0/endpoints/fs.py` the `upload_file` also expects `UploadFile`, so shouldn't `BytesIO` work for uploading a file as well?

scottwn (approved):

Issue #221, opened 2022-03-24
Don't create a new object for every result update


scottwn:
Currently there is no way for the data service to replace or delete an object. Because of this, every update to a results object uploads a new object and then points the dispatch id to that new object, creating many orphaned objects. This could be fixed by one of the following approaches.

* enhance data service to take a put update with a file body so at least we're not doing 2way communication on every update
* enhance data service upload to take a overwrite queryparam
* keep results in a DB somehow?

cjao:
[PR #271](https://github.com/AgnostiqHQ/covalent/pull/271) will grow the ability to delete objects via the data service. Also, the data service would be able to replace an object if not for [these lines](https://github.com/AgnostiqHQ/covalent/blob/develop/refactor/data/app/core/localstoragebackend.py#L110-L112).

PR #271, opened 2022-03-31
Add delete endpoint to object and results services

cjao:

Closes https://github.com/AgnostiqHQ/covalent-staging/issues/432


refactor/results/app/api/api_v0/endpoints/workflow.py
```diff
@@ -217,3 +217,15 @@ def update_result(*, dispatch_id: str, task: Node) -> Any:
...
         return {"response": "Task updated successfully"}
+
+
+@router.delete("/results/{dispatch_id}", status_code=200)
+def delete_result(*, dispatch_id: str) -> Any:
+    # Retrieve file path from db and request deletion from data service
+    filename = _get_result_from_db(dispatch_id, "filename")
+    r = requests.delete(f"http://{base_url}/delete", params={"obj_name": filename})
+
+    sql = "DELETE FROM results WHERE dispatch_id = ?"
```
scottwn, line 228: This should only delete from the DB if the object deletion was successful. Otherwise we will create orphaned objects.

scottwn (changes requested):
Do you intend for this to fix the issue in #221? I think the problem will remain even after merging this

cjao:
> Do you intend for this to fix the issue in #221? I think the problem will remain even after merging this

#272 would fix #221 unless you see some problem with that approach.

cjao:
Any more changes needed?

scottwn (approved):

PR #272, opened 2022-03-31
Allow replacing objects in LocalStorageBackend

cjao:

This would fix https://github.com/agnostiqhq/covalent/issues/221.


scottwn:
this is an alternate approach to #278 
@wjcunningham7 @kessler-frost can you decide which to merge and which to close?

kessler-frost:
I like the 278 approach better as we can decide whether to overwrite or not whereas in this approach it is always going to overwrite if object already exists.

PR #278, opened 2022-03-31
overwrite objects in data service

scottwn:


refactor/data/app/core/localstoragebackend.py
```diff
@@ -108,7 +109,7 @@ def put(
             return ("", "")
         try:
             abs_p = str(p.resolve(True))
-            object_name = f"{uuid.uuid4()}.pkl"
+            object_name = f"{uuid.uuid4()}.pkl" if not overwrite else object_name
```
kessler-frost, line 112: Do you think it'll be better to use `tempfile._get_candidate_names()` instead of `uuid` here?

refactor/data/app/api/api_v0/endpoints/fs.py
```diff
@@ -58,7 +58,7 @@
 
 
 @router.post("/upload", status_code=200, response_model=UploadResponse)
-def upload_file(*, file: UploadFile) -> Any:
+def upload_file(*, file: UploadFile, overwrite: bool = False) -> Any:
```
kessler-frost, line 61: I also think that by default `overwrite` should be `True`.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+### Fixed
+
+- Using functions for getting result object in cancel endpoint and sending cancel task signal to runner in the dispatcher.
+
```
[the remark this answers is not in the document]
scottwn: huh, I didn't write that comment.... I wonder how that got there. I'll fix it.

refactor/data/app/api/api_v0/endpoints/fs.py
scottwn: I think the default should be to _not_ delete files. User should have to send an extra param to destroy data.

kessler-frost: Oh, sorry, you're write. I meant that we need to pass `overwrite=True` when the result service is interacting with data service. So every time result calls this endpoint, overwrite should always be done.

refactor/data/app/core/localstoragebackend.py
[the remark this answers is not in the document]
kessler-frost: Yeah.

cjao: This doesn't work for me. `tempfile._get_candidate_names()` returns a `tempfile._RandomNameSequence object`, so

`f"{tempfile._get_candidate_names()}.pkl"` evaluates to `'<tempfile._RandomNameSequence object at 0x7f40363062b0>.pkl'
`.

The current code works.

kessler-frost: Yeah, you're right. I was just wondering if there was a more standard/specific way of generating file names without creating files but it seems there isn't. So, yeah, `uuid` should work fine.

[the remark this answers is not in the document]
kessler-frost: No longer needs to be changed.

cjao (approved):

kessler-frost (approved):

PR #282, opened 2022-03-31
Stop creation of a pickled result file on client side

### LINKED ISSUE #256: Don't create local results file
Today, when you do `ct.dispatch()`, there is a pickle file created on the client side and then uploaded to the queuer. But when the dispatcher updates the results, that happens in the results service and the client side results file is not updated, thus causing a divergence problem.

The solution is to not create the results pickle file on the client side and instead send bytes to the queuer. 

There should also be a `ct.get_results()` function that returns the result file as bytes.

Related to #242

kessler-frost:

Fixes #256.


refactor/client_side/interface_with_covalent.py
```diff
@@ -0,0 +1,77 @@
...
+from functools import wraps
+from io import BytesIO
+
+import cloudpickle as pickle
+import requests
+
+from covalent._results_manager.result import Result
+from covalent._workflow.lattice import Lattice
+
+SUBMIT_DISPATCH_QUEUER_ADDR = "http://localhost:8000/api/v0/submit/dispatch"
```
FyzHsn, line 36: Use the Settings class and `get_route` method. No more hardcoded file paths.

refactor/client_side/interface_with_covalent.py
```diff
@@ -0,0 +1,77 @@
...
+GET_RESULT_ADDR = "http://localhost:8000//api/v0/workflow/results"
```
FyzHsn, line 37: Same as above.

FyzHsn (changes requested):

refactor/client_side/interface_with_covalent.py
kessler-frost: Updated.

refactor/client_side/interface_with_covalent.py
kessler-frost: Updated.

PR #307, opened 2022-04-05
supporting batch delete

wjcunningham7:


wjcunningham7:
Major notes:
- The delete endpoints in both the results and data services support list query strings
- The response for both return information about what succeeded and what failed
- The local storage backend has a similar response structure
- The SQL commands performed in the result service's delete endpoint can be improved; currently they query/delete one-by-one

wjcunningham7:
@AlejandroEsquivel again I'm not sure why a major portion of the spec was deleted

refactor/results/app/api/api_v0/endpoints/workflow.py
```diff
@@ -274,16 +278,32 @@ def update_result(*, dispatch_id: str, task: bytes = File(...)) -> Any:
         return {"response": "Task updated successfully"}
 
 
-@router.delete("/results/{dispatch_id}", status_code=200)
-def delete_result(*, dispatch_id: str) -> Any:
-    # Retrieve file path from db and request deletion from data service
-    filename = _get_result_from_db(dispatch_id, "filename")
-    r = requests.delete(f"http://{base_url}/delete", params={"obj_name": filename})
+@router.delete("/results", status_code=200, response_model=DeleteResultResponse)
+def delete_result(*, dispatch_ids: List[str] = Query([])) -> DeleteResultResponse:
```
valkostadinov, line 282: `List[str]` will expect the query to be of the form: `?dispatch_ids=xxx&dispatch_ids=xxx&dispatch_ids=xxx` which is verbose and redundant but it works. A comma-delimited list seems better here but it seems to require a customer query parser so probably not worth the effort.

valkostadinov (dismissed):

PR #308, opened 2022-04-05
Updating the actions to push microservice images to public repos.

scottwn:
This commit also contains lots of style changes due to `pre-commit run
--all-files`.

I can split those changes into another PR if it's too difficult to merge
them now.

AgnostiqHQ/covalent-infra#18


Emmanuel289 (commented):
Should we assign DATA_SVC_PORT to a different number than 8004 since the UI service is using both 8004 and 8005?

scottwn:
In the cloud environment the port doesn't matter.... maybe we should assign them all to `80`?

Emmanuel289:
> In the cloud environment the port doesn't matter.... maybe we should assign them all to `80`?

Okay, that works.

.github/workflows/publish_develop.yml
```diff
@@ -25,8 +25,8 @@ on:
     branches:
       - develop
     paths-ignore:
-      - 'CHANGELOG.md'
-      - 'VERSION'
+      - "CHANGELOG.md"
```
wjcunningham7, line 28: does single vs double quote matter in the actions?

scottwn: It does if you're doing variable expansion or using some related shell commands, but it doesn't matter here. This is enforcing the precommit rule.

scottwn:
@Emmanuel289 @venkatBala I'm thinking more about the right way to do the port mapping in the docker compose files. I was thinking that we can assign everything to the same port in `docker-compose.yml` since each service will be running on a separate address. But I guess that would break it if you wanted to run it locally? But then again... maybe the purpose of `docker-compose.yml` isn't for local hosting? What do you think?

.github/workflows/publish_develop.yml
```diff
@@ -69,3 +69,20 @@ jobs:
...
+          docker tag covalent-queuer:latest ${{ secrets.QUEUER_DEV_REPO }}
+          docker tag covalent-results:latest ${{ secrets.RESULTS_DEV_REPO }}
+          docker tag covalent-runner:latest ${{ secrets.RUNNER_DEV_REPO }}
+          docker tag covalent-ui:latest ${{ secrets.UI_DEV_REPO }}
+          docker push ${{ secrets.DATA_DEV_REPO }}
+          docker push ${{ secrets.DISPATCHER_DEV_REPO }}
+          docker push ${{ secrets.QUEUER_DEV_REPO }}
+          docker push ${{ secrets.RESULTS_DEV_REPO }}
+          docker push ${{ secrets.RUNNER_DEV_REPO }}
+          docker push ${{ secrets.UI_DEV_REPO }}
```
venkatBala, line 88: should we use secrets for the public ECR registries?

scottwn: That's the standard that we've been using. Obviously the repos themselves are not secret, but we have put all the ECR repos into secrets for consistency.

venkatBala: sounds good!

scottwn:
I guess we can't assign both the backend and the frontend on the UI to port `80`.... how should we map the UI ports?

Emmanuel289:
> @Emmanuel289 @venkatBala I'm thinking more about the right way to do the port mapping in the docker compose files. I was thinking that we can assign everything to the same port in `docker-compose.yml` since each service will be running on a separate address. But I guess that would break it if you wanted to run it locally? But then again... maybe the purpose of `docker-compose.yml` isn't for local hosting? What do you think?

I was thinking about this point as well. `docker-compose.yml` needs unique port allocations for local hosting. Maybe supervisord is a better alternative for running the services locally

scottwn:
yeah I think `docker-compose.yml` should not be designed to run locally

Emmanuel289:
> I guess we can't assign both the backend and the frontend on the UI to port `80`.... how should we map the UI ports?

80 and 8080?

scottwn:
which should be which? `80:8004` and `8000:8005`?

Emmanuel289:
> which should be which? `80:8004` and `8080:8005`?

Yes, these should be good

scottwn:
We removed `8005` because it seemed like it wasn't doing anything. Was that a mistake?

docker-compose.yml
```diff
@@ -6,62 +6,58 @@ version: "3.9"
...
   covalent-ui:
-    image: "${UI_REPO}"
+    image: "public.ecr.aws/covalent/covalent_ui"
     ports:
-      - 3001:8004
-      - 8005:8005
-    environment:
-      UI_SVC_HOST: 0.0.0.0
-      UI_SVC_PORT: 8004
+      - "80:8004"
```
scottwn, line 60: @Emmanuel289 can you confirm that this piece looks OK from your side?

Emmanuel289: yeah, this looks good

PR #316, opened 2022-04-06
Fix UI docker connection issue + update dockerfiles

venkatBala:


Emmanuel289 (changes requested):
@venkatBala. We don't need to update VERSION in our PRs now, because the CovalentOps Bot now takes care of that (and version workflow actually fails if we try to update VERSION)

VERSION
```diff
@@ -1 +1 @@
-0.61.0
+0.62.0
```
Emmanuel289: We don't need to update VERSION now, because that's now automated.

scottwn (approved):

Emmanuel289 (approved):

scottwn (approved):

PR #321, opened 2022-04-06
Docker issues

scottwn:


docker-compose.yml
```diff
@@ -1,63 +1,91 @@
...
   covalent-queue-consumer:
-    image: "public.ecr.aws/covalent/covalent_queuer"
+    build:
+      context: .
+      dockerfile: refactor/queuer/Dockerfile
+    image: covalent-queue-consumer
+    command: sh -c "python main.py"
     ports:
-      - "80:8001"
+      - "3000:8001"
```
wjcunningham7, line 52: does this service need to expose a port?

docker-compose.yml
```diff
@@ -1,63 +1,91 @@
...
   covalent-ui:
-    image: "public.ecr.aws/covalent/covalent_ui"
+    build:
+      context: .
+      dockerfile: refactor/ui_backend/Dockerfile
+    image: covalent-ui
+    command: sh -c "python main.py"
     ports:
-      - "80:8004"
+      - "8005:8004"
```
wjcunningham7, line 88: should this be 8005:8005?

wjcunningham7 (approved):

docker-compose.yml
scottwn: no, the default port for the UI backend is `8004`....... I think 🙃 
supervisord sets it to `8005`, so docker-compose needs to do the same thing

docker-compose.yml
scottwn: I'm not sure... @AlejandroEsquivel @Emmanuel289 does this look right for the consumer?

Emmanuel289: @scottwn I exposed port 8001, similar to the queuer. Maybe @AlejandroEsquivel can confirm if this is correct for the service.

AlejandroEsquivel: @scottwn @Emmanuel289 the queue consumer does not expose an API, it just connects to NATS so we do not require port mappings here

FyzHsn (approved):

PR #337, opened 2022-04-08
Configuration Normalization using root level .env file

AlejandroEsquivel:

### Changes

- Now `.env` file is not checked in to git, `covalent/.env.example` is effectively the template, and it is copied into `.env` when running covalent (as this file is mutated)

### CLI

You can set covalent configuration by using the cli command:
```shell
# legacy config
covalent config legacy_dispatcher.port=48008
# new config
covalent config MQ_CONNECTION_URI=localhost:4200
covalent config DISPATCHER_SVC_HOST=localhost DISPATCHER_SVC_PORT=8002
```

You can introspect covalent configuration by using:
```shell
covalent config
```


covalent/_shared_files/config.py
```diff
@@ -24,12 +24,53 @@
...
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, MutableMapping, Optional, Union
 
 import toml
+from dotenv import dotenv_values, find_dotenv, set_key
 
 """Configuration manager."""
 
+CONFIG_FILE_NAME = ".env"
+PROJECT_ROOT = os.path.dirname(os.path.abspath(__file__)) + "/../.."
```
wjcunningham7, line 35: if the package is pip installed, this will point to `site-packages`... is that what we want?

refactor/results/app/core/config.py
```diff
@@ -46,6 +46,10 @@ def assemble_cors_origins(cls, v: Union[str, List[str]]) -> Union[List[str], str
 
     class Config:
         case_sensitive = True
+        env_file = ".env"
 
 
 settings = Settings()
+
+print("Configuration:")
```
wjcunningham7, line 54: are these debugging statements that should be removed?

refactor/ui_backend/app/core/config.py
```diff
@@ -41,6 +43,10 @@ def assemble_cors_origins(cls, v: Union[str, List[str]]) -> Union[List[str], str
 
     class Config:
         case_sensitive = True
+        env_file = ".env"
 
 
 settings = Settings()
+
+print("Configuration:")
```
wjcunningham7, line 51: same comment

covalent/_shared_files/config.py
AlejandroEsquivel: Given the config file is not intended to be edited directly by the user I think this suffices, I cannot think of other negative side effects of this as of now however.

covalent_dispatcher/_cli/service.py
```diff
@@ -41,9 +41,11 @@
 UI_PIDFILE = get_config("dispatcher.cache_dir") + "/ui.pid"
 UI_LOGFILE = get_config("user_interface.log_dir") + "/covalent_ui.log"
 UI_SRVDIR = os.path.dirname(os.path.abspath(__file__)) + "/../../covalent_ui"
-SD_PIDFILE = os.path.dirname(os.path.abspath(__file__)) + "/../../supervisord.pid"
 SD_CONFIG_FILE = os.path.dirname(os.path.abspath(__file__)) + "/../../supervisord.conf"
```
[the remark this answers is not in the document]
AlejandroEsquivel: Same comment

refactor/results/app/core/config.py
AlejandroEsquivel: I've found in past projects there is a lot of value displaying app configuration when spinning up a service, so it is intended to stay there

refactor/ui_backend/app/core/config.py
AlejandroEsquivel: same comment as above

refactor/runner/app/core/config.py
```diff
@@ -44,6 +47,10 @@ def assemble_cors_origins(cls, v: Union[str, List[str]]) -> Union[List[str], str
 
     class Config:
         case_sensitive = True
+        env_file = ".env"
```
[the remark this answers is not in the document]
AlejandroEsquivel: same comment as above

.env.example
```diff
@@ -0,0 +1,52 @@
+#[Microservices Configuration]
```
wjcunningham7, line 1: this file seems to not get packaged properly.  all files that get installed to site-packages need to reside inside one of the package subdirectories.

wjcunningham7: to reproduce the error:
```
python setup.py sdist
pip install dist/*.tar.gz
covalent start --refactor
```

AlejandroEsquivel: should be able to start covalent now

kessler-frost (approved):

PR #363, opened 2022-04-13
nats installation using pip

wjcunningham7:


wjcunningham7:
remaining:
- error handling
- check if nats-server already is installed before including
- investigate role of `pyproject.toml` file
- cleanup and documentation

wjcunningham7:
nats-server is only installed when installing via pip from pypi.  To mimic this behavior:
```
python setup.py sdist
pip install dist/*.tar.gz
```

alternatively you can run one of 
```
pip install .
python setup.py install
```

If things are installed correctly, you should be able to run `nats-server` from the command line.

Note: If you run `pip install -e .` it will not install nats.

wjcunningham7:
testing this in isolation with `supervisord`:

```
supervisorctl start covalent:nats
```

scottwn (approved):

PR #364, opened 2022-04-13
Update/fix the Dockerfiles to use multi-stage builds to reduce container image sizes

venkatBala:


scottwn (changes requested):
- no changes to CHANGELOG outside of the UNRELEASED block
- no changes to VERSION
- example_dispatch.py should not be in the root
- scikit should not be in requirements

scottwn:
pre-commit.ci autofix

scottwn (dismissed):

PR #366, opened 2022-04-13
Some fixes for the star tracker tutorial

kessler-frost:

Resolves https://github.com/agnostiqhq/covalent-staging/issues/455 .


refactor/client_side/example_dispatch.py
```diff
@@ -55,9 +56,9 @@ def workflow(a):
 print(result)
 
 
-print("Now using dispatch_sync: ")
+# print("Now using dispatch_sync: ")
 
-print(ct.dispatch_sync(workflow)(2))
+# print(ct.dispatch_sync(workflow)(2))
 # time.sleep(3)
```
scottwn, line 62: comments should be just a comment instead of a commented print statement

scottwn (changes requested):
just a small change, and please double check that you intend for the print statements in update_workflow.py to be there.... I don't know how logging works in the dispatcher, double checking that those get sent to the logs

refactor/client_side/example_dispatch.py
kessler-frost: Yeah, they go to the supervisord logs, since it basically captures all the stdout and sends it to a file.

kessler-frost: Fixed this comment.

scottwn (approved):

PR #384, opened 2022-04-17
fixing the nats shutdown

wjcunningham7:


wjcunningham7:
Tested on Linux but not on Mac

wjcunningham7:
summary of the bug:
- `nats-server` is installed as a console script by `setuptools`, which in turn means the binary `nats-server` is invoked as a Python subprocess
- `supervisord` manages the parent process and therefore kills the parent, not the child
- sending a `SIGINT` signal to the parent also kills the child, but other signals like `SIGTERM`, `SIGHUP`, and `SIGKILL` do not

PR #395, opened 2022-04-18
Fix for unpickling result object in different environment

### LINKED ISSUE #313: Import package installation inside electron's container
### What should we add?

Currently although we collect all of the imports (and more) for each lattice that we are constructing, we still do not collect their name space installation in pip server. This is needed if we are installing the said packages, wherever we are running the electrons. Interestingly it turns out to be quite straight forward to do that. (This becomes very important when name space and import name do not match, like covalent-cova). Using this, one can directly create the requirements.txt file for any given lattice/dispatch. 

```python
from importlib import metadata
from collections import defaultdict
dists = metadata.distributions()
pkg_to_dist = defaultdict(list)
for dist in dists:
    if dist.read_text('top_level.txt') is not None:
        top_level=dist.read_text('top_level.txt').split()
        for pkg in top_level:
            pkg_to_dist[pkg]=dist.metadata['Name']
print(pkg_to_dist['covalent'])
```
```
cova
```

_No response_

kessler-frost:

Services such as the Results service currently need the packages installed in the user's environment, to be installed in its own environment so that it can update or insert the result object in its database since it needs to unpickle the result object in its entirety. This PR solves that issue by only unpickling the parts which are actually needed to be updated and not the whole object, which means that the service which doesn't actually call the function will not need to have the users packages installed in its python environment.

Resolves #313.


kessler-frost:
Currently the UI fails in this branch due to some changes on how pickling and unpickling is done. Working on fixing that too.

scottwn (approved):

scottwn (dismissed):

kessler-frost:
Updating the `tests_assets` of multiple services led to the discovery of this issue with results service's PUT test: https://github.com/agnostiqhq/covalent/issues/407 in develop branch. I will be skipping that test in this branch since it is not really a blocking issue for this PR.

wjcunningham7 (approved):

PR #397, opened 2022-04-18
actions cleanup

scottwn:


scottwn:
Need to edit publish and publish_dev so that they are only building the service containers

scottwn:
TODO: Fix the version tagging for the service images in ECR. Will be done as part of AgnostiqHQ/covalent-infra#23

.github/workflows/release.yml
```diff
@@ -57,20 +82,93 @@ jobs:
...
+        run: |
+          if [ ${{ github.event.inputs.prerelease }} = 'false' ] && \
+             [ -z ${{ github.event.inputs.stable_version }} ] ; then
+            echo "You can't create a stable release without specifying the stable version number."
+            exit 1
+          fi
+          VERSION="$(cat ./VERSION)"
+          cd dist
+          tar xzf cova-${VERSION}rc0.tar.gz
+          diff -x tests -r cova-${VERSION}rc0/covalent ../covalent
```
wjcunningham7, line 151: the name of the tarball will depend on whether it's a pre-release or stable release

wjcunningham7 (changes requested):

.github/workflows/release.yml
scottwn: oh right of course, thanks for catching that

wjcunningham7 (approved):

PR #411, opened 2022-04-19
properly update workflow tasks instead of pushing new results

### LINKED ISSUE #409: Update workflow logic leads to race condition
### Environment

- **Covalent version**: 0.88.0
- **Python version**: 3.8.13
- **Operating system**: Gentoo Linux

### What is happening?

Race condition happens in the QG tutorial and sometimes elsewhere.  I suspect this is due to logic in the file `covalent_dispatcher/app/core/update_workflow.py` not properly calling the results service to update the status of a completed task.  Currently we use `send_result_object_to_result_service` everywhere, which calls the service's `POST` method, but we should also be using `send_task_update_to_result_service` which does a read/modify/write operation directly in the results service.

Run the QG tutorial

### What should happen?

Tutorial runs smoothly

### Any suggestions?

see above

scottwn:

Resolves #409


FyzHsn (dismissed):

FyzHsn (approved):

scottwn:
@FyzHsn is there an issue open to re-enable the skipped tests?

PR #424, opened 2022-04-19
Updated .env location to $HOME/.cache/covalent. Updated CHANGELOG

AlejandroEsquivel:
- [ ] I have added the tests to cover my changes.
- [x] I have updated the documentation and CHANGELOG accordingly.
- [x] I have read the CONTRIBUTING document.

PR #429, opened 2022-04-20
dotenv hotfix

wjcunningham7:

### DIFF
--- a/covalent/_shared_files/config.py
+++ b/covalent/_shared_files/config.py
@@ -79,36 +79,17 @@ class _ConfigManager:
     """
 
     def __init__(self) -> None:
-
-        # ensure .env exists, if not copy from .env.example
         self.env_path = os.environ.get("ENV_DEST_DIR") or PROJECT_ROOT
+        self.config_file = os.path.join(self.env_path, CONFIG_FILE_NAME)
+        # ensure .env exists, if not copy from .env.example
         self.ensure_config_file_exists()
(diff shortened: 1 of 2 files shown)


wjcunningham7:
@scottwn this is a hotfix for the stable release

covalent/_shared_files/config.py
```diff
@@ -79,36 +79,17 @@ class _ConfigManager:
...
-
-        # ensure .env exists, if not copy from .env.example
         self.env_path = os.environ.get("ENV_DEST_DIR") or PROJECT_ROOT
+        self.config_file = os.path.join(self.env_path, CONFIG_FILE_NAME)
+        # ensure .env exists, if not copy from .env.example
         self.ensure_config_file_exists()
-        env_file_path = os.path.join(self.env_path, CONFIG_FILE_NAME)
-
-        # load .env values into os.environ or use explicitly set environment vars
-        load_dotenv(CONFIG_FILE_NAME)
```
AlejandroEsquivel, line 89: This can be removed without any repercussions, however for consistency we may want to have config management to work in the same way as the microservices so that the variables defined in the `.env` file are also accessible via environment variables which can only happen if the `load_dotenv` statement remains as: `load_dotenv(self.config_file)`. However this is addressed in PR #424  so not a blocker here

AlejandroEsquivel (approved):

santoshkumarradha (dismissed):

scottwn (approved):

PR #430, opened 2022-04-20
Make TransportGraph a MultiDiGraph

### LINKED ISSUE #392: Missing positional arguments error thrown when passing same argument twice to electrons in a workflow
**Issue**

When creating a simply workflow, graph build does not recognize the input arguments to an electron when they are exactly the same.

**How to reproduce**
Execute this simple workflow

```python
import covalent as ct

@ct.electron
def add(x, y):
    return x + y

@ct.electron
def multiply(x, y):
    return x*y

@ct.lattice
def workflow(x, y):
    r1 = add(x, y)
    r2 = multiply(r1, r1)
    return r1 + r2

id = ct.dispatch(workflow)(2, 3)
res = ct.get_result(id,wait=True)
```

**Output**

cjao:
The previous DiGraph structure cannot represent the scenario where the
same input node is passed to multiple arguments of an Electron.

Closes #392


cjao:
@wjcunningham7 @kessler-frost @valkostadinov This should fix the back end. Do you want to fix the front end in this branch or pursue that later?

cjao:
Tagging @santoshkumarradha as well.

wjcunningham7 (approved):

wjcunningham7:
no longer needed after migrations

PR #523, opened 2022-05-12
Improve concurrency

### LINKED ISSUE #513: Concurreny issue when scheduling tasks in `execution.py`
Currently when we do a `loop.create_task` in `execution.py` we essentially end up executing each electron sequentially since all of our executors are synchronous.

We might want to rethink whether we want these electrons to be executed in parallel - using multiprocessing, or even multithreading, or we want to do a "submit" of jobs and retrieval of result later - using asyncio.

Both the solutions have downsides - multiprocessing will create a separate process which is not worth it if your function's execution is going to happen on a different machine, and asyncio will need each `executor.execute` function to be async aware - which means changing all the executor plugins we've created and the ones that the user will create (which I don't think is a desirable way).

Using multithreading may be our best bet but it also comes with problems of race conditions and such, we may want to have a longer discussion in how we want things to be.

kessler-frost:
Resolves #513.


covalent_dispatcher/entry_point.py
```diff
@@ -25,6 +25,7 @@
 import sys
 import threading
 import uuid
+from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
```
venkatBala, line 28: Can you remove importing the ProcessPoolExecutor as its not being used?

kessler-frost: Done.

venkatBala:
> Resolves #513 .

lets leave https://github.com/AgnostiqHQ/covalent/issues/513 open... since its not really known for certain that multiprocessing/multi-threading is not a viable alternative for many simultaneous users when self-hosting. Going to need more testing in upcoming sprints.

kessler-frost:
100% Agree, just wanted to connect this PR to that issue for keeping track somehow, since we've had situations before where we couldn't pin point the commit when a feature was introduced.

kessler-frost:
We can merge this and reopen that issue afterwards.

venkatBala:
> We can merge this and reopen that issue afterwards. If that works?

Awesome... sounds good!

kessler-frost:
Because it is still partially resolving that issue with the current `develop-covalent`.

Issue #547, opened 2022-05-19
Covalent status is unreliable


cjao:
### Environment

- **Covalent version**: 0.33.1
- **Python version**: 3.8.13
- **Operating system**: Fedora 35

### What is happening?

`covalent status` can report that the server is still running even after I manually kill the server process or the server crashes.

### How can we reproduce the issue?

1. `covalent start`
2. Find the pid of the server process via `ps aux | grep python app.py`, and kill that pid. Confirm that `ps` shows no covalent server running.
3. `covalent status` yields `Covalent server is running at http://0.0.0.0:48008.`

### What should happen?

`covalent status` should report the server as stopped.

PR #579, opened 2022-05-31
504 default dask cluster startup when no arguments provided

Emmanuel289:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 cloudpickle==2.0.0
+dask[distributed]==2022.1.0
 eventlet==0.31.0
 flask==2.0.1
 flask-cors==3.0.10
--- a/covalent/_shared_files/defaults.py
+++ b/covalent/_shared_files/defaults.py
@@ -56,6 +56,14 @@ _DEFAULT_CONFIG = {
         "log_dir": (os.environ.get("XDG_CACHE_HOME") or (os.environ["HOME"] + "/.cache"))
         + "/covalent",
     },
+    "dask": {
+        "scheduler_address": os.environ.get("scheduler.address"),
+        "scheduler_port": os.environ.get("scheduler.port"),
+        "cache_dir": (os.environ.get("XDG_CACHE_HOME") or (os.environ["HOME"] + "/.cache"))
+                     + "/covalent",
+        "log_dir": (os.environ.get("XDG_CACHE_HOME") or (os.environ["HOME"] + "/.cache"))
+                   + "/covalent",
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -1,4 +1,5 @@
 cloudpickle==2.0.0
+dask[distributed]==2022.1.0
```
santoshkumarradha, line 2: Hey @Emmanuel289 , quick comment, instead of this, please make our dask executer plugin a requirement, which intern has dask and other things.

santoshkumarradha, line 2: Hey @Emmanuel289 , quick comment, instead of this, please make our dask executer plugin a requirement, which intern has dask and other things.

covalent_dispatcher/_cli/service.py
```diff
@@ -135,6 +138,18 @@ def _is_server_running() -> bool:
...
 
 
+def _is_dask_running() -> bool:
+    """Check status of the Dask cluster's scheduler.
+
+    Returns:
+        status: Status of whether the scheduler is running
+
+    """
+    if _read_pid(DASK_PIDFILE) == -1:
```
cjao, line 148: This is susceptible to same [problem](https://github.com/AgnostiqHQ/covalent/issues/547) affecting `covalent status`. If the dask process crashes the PID file won't be cleaned up. Besides checking whether the pidfile exists we should also check whether the process is still alive.

Emmanuel289: Thanks @cjao. I'll add a check for the process liveness

covalent_dispatcher/_cli/service.py
```diff
@@ -179,6 +195,46 @@ def _graceful_start(
...
+    if psutil.pid_exists(pid):
+        addr = get_config("dask.scheduler_address")
+        port = get_config("dask.scheduler_port")
+        click.echo(f"Dask scheduler is already running at {addr}.")
+        return addr, port
+
+    _rm_pid_file(pidfile)
+    pypath = f"PYTHONPATH={UI_SRVDIR}/../tests:$PYTHONPATH" if develop else ""
+    launch_str = f"{pypath} python app_dask.py >> {logfile} 2>&1"
+    proc = Popen(launch_str, shell=True, stdout=DEVNULL, stderr=DEVNULL, cwd=server_root)
```
venkatBala, line 229: instead of routing the `stdout` and `stderr` to DEVNULL why not use PIPE, this way we can communicate with the process to check `stdout` and `stderr` logs

PR #609, opened 2022-06-06
Simpler dask startup

venkatBala:


covalent_ui/app.py
```diff
@@ -43,6 +47,38 @@
...
+            {
+                "dask": {
+                    "scheduler_address": scheduler_address,
+                    "scheduler_port": scheduler_port,
+                }
+            }
+        )
+
+        # Halt the process here until its terminated
+        signal.pause()
```
kessler-frost, line 79: I don't think we should split the address into address and port since in the case of remote clusters there will be no port, and splitting it does not give us any advantage either afaik.

kessler-frost (changes requested):

scottwn (dismissed):

covalent_ui/app.py
venkatBala: correct... I did not make any changes to the original code in `start_cluster`.. we can update it here

CHANGELOG.md
```diff
@@ -7,6 +7,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Changed
+
+- Skipping stdout/stderr redirection tests until implemented in Dask parent process
+
+### Added
+
+- Simplifed starting the dask cluster using `multiprocessing`
+- Added `bokeh==2.4.3` to requirements.txt to enable view Dask dashboard
+
+## Fixed
```
cjao, line 19: Could you fix the header here? This was from an earlier commit.

venkatBala: fixed

PR #614, opened 2022-06-07
612 add Ensemble_pennylane notebook

poojithurao:


AnnaGwen:
Looks good, Poojith! I have a handful of changes that I’d like you to make

### **Structural Changes**

1. Overall wording
We should include a link to the original Pennylane tutorial, but the wording in our tutorial should be different. It can be similar since it is describing the same basic concepts, but it can’t be identical to what they have.
This tutorial would also benefit from more and longer text descriptions in general.

2. The draw() function
Thanks for starting the issue on this! We can work around this by removing the draw portion of the tutorial altogether. Instead, take a screenshot of the DAG that gets generated by the UI when the workflow is dispatched, and incorporate that into the tutorial.

3. The devices
Since we’re getting errors related to the qml devices, we should have a different approach for now. Rather than combining results from two different devices, we should combine results using different input parameters (as you have in the code now). This will mean changing the text descriptions to reflect that this is actually what we’re doing, as well as removing parts of the code where the devices are specified.

4. Conclusion
We’ll need some final paragraph or two summarizing what the results are showing us. You can do a visualization again, just leave it outside the covalent workflow.

### **Wording Changes**

* I know they don’t have this in the Pennylane tutorial, but can you add a short description about the Iris dataset? I.e. what the features are that we’re looking at, and how we’re trying to classify them.

* Under the ‘Plot the Data’ you can note that we are doing this for visualization purposes but it is not included in the workflow

* functions with ct.electron decorator → functions with the Covalent electron decorator

* Missing a description of the decision() function

* with the ct.lattice decorator → with the Covalent lattice decorator

### **Typos**

* acccuracy -> accuracy 

* In code block starting with line [57] there is a commented ​​#p_train that can be removed

* we can submit the task via the dispatch method → we can submit the workflow using Covalent dispatch

AnnaGwen:
This is looking a lot better with additional text. I just have a few more minor wording changes and then we should be good to go. 
___


- In the title, ‘ensemble classification’ should be capitalized

- In this tutorial we show how a quantum circuit → In this tutorial, we show how a quantum circuit  

- Make sure to capitalize Covalent throughout

- Before starting the first code block, include text saying something like “We start by importing all the necessary packages.”

- split them into training and test dataset → split them into a training and test dataset

- Iris dataset is a very well known dataset used for pattern recognition  → The Iris dataset is very well known in machine learning, and used for pattern recognition 

- 3 classes namely: → 3 classes:

- using PCA → using principal component analysis

- We plot the data to visualize the data → We plot the data to visualize it

- Transport graph → Directed Acyclic Graph (DAG)

- we can submit the workflow using Covalent dispatch → we can submit it using Covalent dispatch

- using the ct.get_result method → using the Covalent get_result method

- In order to calculate how well the ensemble model performs we use the predictions to calculate the accuracy of the model. → In order to calculate how well the ensemble model performs, we compare the model predictions to the actual classification values.

-  test set and it can be observed that the ensemble model performs better than the individual circuits →  test set, and it can be observed that the ensemble model performs better than the individual circuits.

- In this tutorial we created workflow using covalent to predict →In this tutorial, we created a Covalent workflow to predict

- Following are the key takeaways from the tutorial. → The key takeaways from this tutorial are:

- Covalent is helpful in creating workflows that can take advantage of parallel execution of quantum circuits → Covalent helps create workflows that can take advantage of parallel execution of quantum circuits and can organize, track, and dispatch experiments.

- The accuracy of ensemble model outperforms predictions from individual parameter settings. → The ensemble model outperforms both individual parameter models.

- We don’t need point 3) in the key takeaways

AnnaGwen (approved):

wjcunningham7 (dismissed):

wjcunningham7 (approved):

PR #627, opened 2022-06-09
Implement the CLI to support starting covalent with/without dask as well as an easy way to quickly parse the logs

venkatBala:


covalent_dispatcher/_cli/service.py
```diff
@@ -325,3 +336,20 @@ def purge() -> None:
     cm.purge_config()
 
     click.echo("Covalent server files have been purged.")
+
+
+@click.command()
+def logs() -> None:
+    """
+    Retrive Covalent logs from file
```
cjao, line 344: "Show Covalent server logs"

covalent_dispatcher/_cli/service.py
```diff
@@ -236,13 +244,16 @@ def _graceful_shutdown(pidfile: str) -> None:
     help="Server port number.",
 )
 @click.option("-d", "--develop", is_flag=True, help="Start the server in developer mode.")
+@click.option("--no-cluster", is_flag=True, help="Start the server without Dask")
```
cjao, line 247: We should mention that this would default the executor to "local".

covalent_ui/app.py
```diff
@@ -172,6 +177,7 @@ def serve(path):
         action="store_true",
         help="Start the server in developer mode.",
     )
+    ap.add_argument("--no-cluster", required=False, help="Start Covalent server without Dask")
```
cjao, line 180: We should mention that this defaults the executor to "local".

cjao (changes requested):

covalent_dispatcher/_cli/service.py
venkatBala: Updated the help menu

PR #634, opened 2022-06-10
Pennylane kernel tutorial

AnnaGwen:


AnnaGwen:
## Structural Changes

1. The integration with Covalent needs some work. There are too many lattices – there should only be one lattice for each workflow. For example, quantum_model_train and quantum_model_predict should be electrons, and the workflow should be constructed from the steps in code block [16].
We should meet to talk about this, but in the meantime take a look at the other tutorials in the Covalent documentation.

2. As before, some of the wording is too similar to the Pennylane tutorial. The wording needs to be unique, and we need to link to the Pennylane tutorial for attribution.

3. The introduction is very short - it would be helpful to expand this and give more background so people can follow along both in the introduction and throughout the tutorial. It doesn’t need to be as long as the Pennylane intro, but it would be good to include an overview of quantum kernels and variational circuits, and why we want to compare them.

4. The comments in the code blocks need to be removed or changed; as is, they are identical to the Pennylane tutorial

5. Provide short explanations of things like n_qubits and dev_kernel

6. The workflows should be under their own headings; right now the lattices get buried with all the electrons

7. There needs to be some explanation about using Covalent dispatch and the get_result method

8. Include a screenshot of the DAG. Be sure to call it a directed acyclic graph (DAG) the first time it is mentioned

9. Add a sentence or two comparing the accuracy scores of the SVM and the variational circuit.

10. The final functions need an explanation as to why they’re not part of the workflow, but you should definitely include them and the plot because they’re crucial for comparing which method scales best with data size.

11. There needs to be a conclusion wrapping up the tutorial and explaining what we’ve learned.

## Wording Changes

Title: Comparing Kernel-Based and Variational Training with Covalent

Before the first code block, you should say something like, “We start by importing all necessary packages.”

using PCA → using principal component analysis

with Covalent lattice decorator for creating the workflow → with Covalent lattice decorator to create the workflow

The workflow runs each step of optimization inside the lattice and hence it can save the intermediate optimization steps. → The workflow runs each step of optimization inside the lattice, and hence it can save the results of intermediate optimization steps.

## Typos

how the results are saved as each step → how the results are saved at each step

## Coding Issues

In code block [16], there’s a problem with the line:
pred_result = ct.get_result(dispatch_id=dispatch_id_pred, wait=True).
It should be:
pred_result = ct.get_result(dispatch_id=dispatch_id_predict, wait=True).

I tried to run this on my machine, and it timed out at step 16. This might go away when the lattice is constructed differently.

venkatBala:
@poojithurao @AnnaGwen is this one done or work in progress?

wjcunningham7:
unfinished work documented in https://github.com/AgnostiqHQ/covalent/issues/868

PR #639, opened 2022-06-10
633 autoencoder tutorial

ruihao-li:
Work in progress. It seems PyTorch optimizer is not working properly with Covalent, regardless of the type of executor. The training loss stays more or less the same with the number of epochs increasing. Similar observation is found in the [MNIST classifier tutorial](https://covalent.readthedocs.io/en/latest/tutorials/machine_learning/mnist_classifier.html) (code block [14]).


santoshkumarradha:
Hey @ruihao-li , great work ! few minor comments,

1. I would add an electron to create the network in iteself, where various parts of network can be parameterized for experimentation , i.e something like this
```python
@ct.electron
def create_linear_ae(layers):
  class LinearAutoencoder(nn.Module):
    ....

@ct.electron
def create_cnn_ae(layers):
  class CNNAutoencoder(nn.Module):
     ...
```

2. I like the idea of comparing two architectures and thats exactly the kind of experiments we want to showcase. For that, I would do the following

```python
@ct.lattice
def experiment(linear_layers,cnn_layers,epochs):
	train_loader = data_loader(batch_size=batch_size_train, train=True)
    test_loader = data_loader(batch_size=batch_size_test, train=False)
    cnn_model=get_cnn_network(cnn_layers)
    linear_model=get_cnn_network(linear_layers)
    losses=[]
    for i in [cnn_model,linear_model]:
    	tmp_loss=[]
	    optimizer = get_optimizer(model=i, lr=lr)
    	for _ in epochs:
    		loss,optimizer=train_one_epoch(i,optimizer) # this pattern of returning the same object back to the function helps in writing iterative loops
    		tmp_loss.append(loss)
    	losses.append(tmp_loss)
    return losses
```

this ensures that both linear/cnn experiments are run in parallel and all the epoch run's losses are saved/cached as they are costly.  The goal is also to parameterize at the lattice/electron level the things you would want to play around and experiment with as you can dispatch multiple such things to covalent in the background at the same time and various experiments would be running.

ruihao-li:
Thanks for the great suggestions @santoshkumarradha! Will keep that in mind until the gradient issue is resolved.

ruihao-li:
Comments from @santoshkumarradha addressed.

jackbaker1001:
Hey @ruihao-li. I've taken a detailed look through this now. Once again, great work. I have a few comments below:

(1) What does the caption "code" mean on the autoencoder architecture figure? It's not explained in the notebook. (I assume it means latent layer or something?)

(2) When you define parts of some library in the notebook, please make these hyperlinks to the API. E.g, things like `linear` and `conv2d` layers from Pytorch.

(3)  Somewhere before code cell [6], can you define mathematically the reconstruction loss as a function of the NN parameters? It will make the following training cell make more sense this way.

(4) In code cell [13], it would be useful to also print out the minimum value of the trainings loss as well as the testing loss. You make comparison between the two in the text but it is difficult to read off the training loss form the figure above. Alternately, add gridlines to the figure above to the minimum training loss can read off easily.

(5) Can you do things a little differently for code cell [13]?  Ideally, we want to be able to compare between (i) the ideal image (ii) the linear AE reconstructed image and (iii) the convolution AE image on the same line for each Epoch. I guess the reason we don't have this currently is because at a given epoch, the batch approximation chooses different images. This should be fixable by passing the appropriate random seed to the data_loader/ whatever is controlling your batch approximation. If this cannot be done, we should at least show for a few images the **final reconstructions** (i.e, at the minimal training loss) side by side for all three panels mentioned above.

Great work!

ruihao-li:
Thanks for the comments @jackbaker1001! 

(1) I think it's a terminology from [coding theory](https://en.wikipedia.org/wiki/Coding_theory), analogous to "code" in the context of cryptography. So yes, it is the latent layer. I will add a sentence to explain it in the text.

(2) I initially did exactly that, but it turns out that when you have hyperlinks in code font, the font does not render properly on RTD. So I gave that up. It's not a big issue in my opinion. But please let me know if you know a way to make it work.

(3) I'm not sure what you have in mind here exactly. For linear AE, we can indeed write down the explicit mathematical expression for the output as a function of the weights and the input, but this is cumbersome as there are four layers. For CNN this is not trivial. Do you mean something simple like $MSE = \frac{1}{n}\sum_i \abs{Y_i(W) - \hat Y_i(W)}$, where $W$ represents the weights in the model?

(4) Yep will do.

(5) Good point! I will give it a try.

jackbaker1001:
@ruihao-li I took a look through again.

For the optimization problem math, your simple formula you gave above is totally fine. We just want it to be obvious what the optimization routine is doing.

After thats in, I will approve this pull.

jackbaker1001 (approved):

PR #658, opened 2022-06-14
Fix training in MNIST tutorial

scottwn:
We now construct the optimizer inside the training electron instead of
passing an externally constructed optimizer to the training electron.

Also temporarily switch to the "local" executor because the Dask
executor doesn't capture stdout/stderr yet.


scottwn:
hotfix-now

scottwn:
hotfix-now

cjao (approved):

venkatBala (commented):
fixed in #652

Issue #580, opened 2022-05-31
Implement server-side dispatch_sync for sublattices


cjao:
Execution of sublattices involves frequent unpickling of the Result object. When the dispatcher encounters a sublattice, it dispatches the sublattice using dispatch_sync from the client SDK, which is implemented as ct.get_result(ct.dispatch(sublattice), wait=True). Unfortunately, get_result(wait=True) constantly polls result.pkl to see if the workflow has completed.

This inefficiency can be addressed independently of the Result object decomposition. Instead of using the SDK, we could implement a server-side _dispatch_sync which blocks until the dispatch has completed without polling any file. For instance, we can simply wait on the future returned by the threadpool executor when dispatching the sublattice workflow.

PR #694, opened 2022-06-20
Support async executors & Async Dask support

### LINKED ISSUE #693: Support async executors
Let's resurrect @kessler-frost's  asyncio approach but while also supporting synchronous executors. Some immediate beneficiaries include:

* Dask is [async aware](https://distributed.dask.org/en/stable/asynchronous.html) and the executor be modified to take advantage of this. 
* Network-based executors (SSH and Slurm)

cjao:
Run _run_planned_workflow in an asyncio event loop.

In _run_task, call executor.executor in a threadpool executor if it is
an ordinary function; otherwise await on it.

Sublattices are also dispatched asynchronously; instead of using the blocking `dispatch_sync`, we wrap the sublattice workflow in an asyncio Task.

Closes #693 and #580.


wjcunningham7:
> All the tests are passing locally except for the cluster cli ones, not sure why.

@kessler-frost fixed by #845

kessler-frost:
Have created an improved version of this PR [here](#896) as things were getting really messy in this one.

PR #700, opened 2022-06-21
Replace topological sort

### LINKED ISSUE #651: More efficient continuous parallel executions of electrons in a lattice (beyond topological sort)
## Overview

Currently, we dispatch electrons parallelly using the topologically sorted graph where each "set" of electrons is being run parallelly. Instead, we can run things faster if we recheck the dependencies each time based on completed electrons. 

Consider the following graph.
output

Topological sorted values are
`[[0, 1, 2], [3, 4], [5, 9]]`

So right now we run things such that `[0,1,2]`  are sent parallelly. We wait for all of them to finish and then we go to `[3,4]` wait for all of them to finish.. etc.

Now consider the scenario where in the first run `[0,2]` nodes are done. We can in theory go on to `4` even though `1` is not completed. We are assuming in the current implementation that all electrons require an equal amount of time and hence all parallelly sent jobs to complete equally at the same time.  

output1

Solution : Algorithm should be changed to the following
- First parallely execute the first topological electrons i.e.`[0,1,2]`
- As you get back result for each electron, run the following depedency function  on the graph where we have marked `G.nodes[completed_electron_node][completed]=True` (set that this node has indeed finished, we can use the result status of this electron to check it as well)
```python
def give_me_next_nodes(G):
    next_set=[]
    for i in G.nodes:
        if G.nodes[i].get('completed'):
            for children in list(nx.dfs_tree(G,i,1))[1:]:
                can_be_added=True
                for parent in list(nx.dfs_tree(G.reverse(), source=children,depth_limit=1).reverse())[1:]:
                    if not G.nodes[parent].get('completed'):can_be_added=False
                if can_be_added and not G.nodes[children].get('completed'):
                    next_set.append(children)
    return np.unique(next_set)
print(give_me_next_nodes(G))
```
- Run the list of nodes generated by `give_me_next_nodes(G)` in parallel and keep doing it.

This will now output `[4]` for a graph with `G[0]` and `G[2]` set to completed. 

Of course, this is just a basic function and in the actual issue, a more robust check for running nodes, etc.. needs to happen. 

_No response_

cjao:

Maintain the following data structures:
* A "pending_deps" dictionary that tracks for each node the number of
parents that have yet to complete. This is initialized to each node's
in_degree.
* A "tasks_queue" containing nodes that are ready to run. This is
initialized with the parentless nodes (the first layer in the previous
"topological sort").

While not all tasks have been processed, the main thread (running
_run_planned_workflow) fetches tasks from the tasks_queue and
submits them to the execution Thread Pool. When each task completes, its
children's pending_deps is decremented by 1; when a child's
pending_deps reaches 0, the child is pushed to the tasks_queue.

Handle task failures and cancellations by pushing a dummy
node_id (currently -1) to the tasks_queue.

Closes #651


scottwn:
@cjao this is decreasing test coverage by 12%. Can you update the tests so that it isn't decreasing coverage? Seems like the offending modules are app.py and transport.py

cjao:
That's odd. This patch touches neither app.py nor transport.py.

scottwn:
It's removing this call to `get_topologically_sorted_graph()`
https://github.com/AgnostiqHQ/covalent/pull/700/files#diff-a99ca6e887d5a6439efcda540e9e2661cdafa9e3a92109f2ac9c390299f794fbL284
I suspect there's some test that was covering that function via the call you removed... that would explain the drop in coverage for transport.py. Is the function spaghetti code now (can it be removed), or does it just need its own unit test?

cjao:
`get_topologically_sorted_graph()` is no longer used with this patch. Let me try removing that function.

scottwn (approved):

scottwn:
> The remaining references to `get_topologically_sorted_graph` are:
> 
>     * `Lattice.check_constraint_specific_sum; Covalent doesn't current support consumable constraints and this code can be safely deleted.
> 
>     * `Result._write_dispatch_to_python_file`; this appears to reconstructs the workflow source from the Transport Graph. This should be modified to not use  `get_topologically_sorted_graph`.

are you going to update these calls in this branch or should we make new issues for this?

cjao:
I am inclined to make new issues for them.

scottwn:
thanks @cjao, please make the new issues and link them back to this PR

scottwn:
@cjao @venkatBala is this ready to go? I see no activity on it in over a week

cjao:
To my knowledge the implementation is correct. However, the details will change once we move from a thread pool back to a single asyncio event loop for managing a workflow (so as to support async exercutors). I would suggest that @venkatBala wait until that change is merged and then make the necessary adjustments to this PR. The implementation will simplify since locks would no longer be needed to avoid concurrency issues.

scottwn:
@venkatBala @cjao what's the status on this? is it good to go?

covalent_dispatcher/_core/execution.py
```diff
@@ -296,133 +337,128 @@ def task_callback(future: Future):
...
+            else:
+                parent = result_object.lattice.transport_graph.get_dependencies(node_id)[0]
+                output = result_object.lattice.transport_graph.get_node_value(parent, "output")
 
-            if node_name.startswith(
-                (subscript_prefix, generator_prefix, parameter_prefix, attr_prefix)
-            ):
-                if node_name.startswith(parameter_prefix):
-                    output = result_object.lattice.transport_graph.get_node_value(node_id, "value")
+                if node_name.startswith(attr_prefix):
```
scottwn, line 376: @cjao can you confirm that you intend for this to be a 3x nested conditional nested inside a while loop?

cjao: @scottwn Yes.

covalent_dispatcher/_core/execution.py
```diff
@@ -281,9 +283,48 @@ def _run_planned_workflow(result_object: Result, thread_pool: ThreadPoolExecutor
...
+        with lock:
+            result_object._update_node(**node_result)
+            with DispatchDB() as db:
+                db.upsert(result_object.dispatch_id, result_object)
+                db.save_db(result_object)
+            result_object.save()
+            result_webhook.send_update(result_object)
+            g = result_object.lattice.transport_graph._graph
+            if node_result["status"] == Result.COMPLETED:
+                for child in g.neighbors(node_result["node_id"]):
```
cjao, line 302: This logic doesn't work if there are multiple edges between two nodes. Each child node appears only once in `g.neighbors()`, whereas we need to decrement by the number of edges between the parent and child, something like

`pending_deps[child] = pending_deps[child] - len(g.edges(node_result["node_id"], child))`

covalent_dispatcher/_core/execution.py
```diff
@@ -281,9 +283,48 @@ def _run_planned_workflow(result_object: Result, thread_pool: ThreadPoolExecutor
...
+                    pending_deps[child] -= 1
```
cjao, line 303: I overlooked the case where there are multiple edges between two nodes. We should have something like
```python
degree = len(g.edges( node_result["node_id"], child ))
pending_deps[child] -= degree
```

cjao (commented):
The algorithm described in the opening post overlooks the case when parent and child have multiple edges. The `g.neighbors(node)` method seems to return each child node only once, regardless of the number of edges.

venkatBala (approved):

venkatBala (approved):

venkatBala (dismissed):

venkatBala (dismissed):

venkatBala (dismissed):

PR #740, opened 2022-06-29
New layout changes and UI graph API optimization

Prasy12:
Layout changes and UI graph API optimization.


mshkanth (approved):

covalent_ui/webapp/src/App.js
```diff
@@ -20,69 +20,88 @@
...
-    socket.on('draw-request', onDrawRequest)
-    return () => {
-      socket.off('draw-request', onDrawRequest)
-    }
-  })
+     socket.on('result-update', onUpdate)
+     return () => {
+       socket.off('result-update', onUpdate)
+     }
+   }, [dispatch])
```
AlejandroEsquivel, line 77: why is dispatch here being used as a dependency to useEffect ?

covalent_ui/webapp/src/App.js
```diff
@@ -20,69 +20,88 @@
...
-  useEffect(() => {
-    const onUpdate = (update) => {
-      dispatch(
-        fetchResult({
-          dispatchId: update.result.dispatch_id,
-          resultsDir: update.result.results_dir,
-        })
-      )
-    }
+   useEffect(() => {
```
AlejandroEsquivel, line 45: This implementation seems to be effectively re-implementing functionality similar to debouncing which is already included in the lodash library (and tested). All we would need to debounce is the on result update callback. If this can be done in a future PR that would be ideal.

AlejandroEsquivel (approved):

Prasy12:
@AlejandroEsquivel Sure we will look into it.

PR #757, opened 2022-07-01
File Transfer Pre/Post Electron execution using CallDeps & Folder class

AlejandroEsquivel:


covalent/_file_transfer/folder.py
```diff
@@ -0,0 +1,22 @@
+from typing import Optional
+
+from covalent._file_transfer.file import File
+
+
+class Folder(File):
+    """
+    Folder class to store components of provided URI including scheme (s3://, file://, ect.) determine if the file is remote,
+    and acts a facade to facilitate filesystem operations. Is a child of the File class which sets is_dir flag to True.
```
scottwn, line 9:
```suggestion
    Folder class to store components of provided URI including scheme (s3://, file://, ect.), determine if the file is remote,
    and act as facade to facilitate filesystem operations. Folder is a child of the File class which sets `is_dir` flag to True.
```

scottwn (changes requested):

PR #742, opened 2022-06-29
Add deps metadata to Leptons

cjao:
- [ ] I have added the tests to cover my changes.
- [x] I have updated the documentation and CHANGELOG accordingly.
- [x] I have read the CONTRIBUTING document.

Closes #741

Issue #741, opened 2022-06-29
Sync electron metadata with leptons


cjao:
Electrons [recently acquired](https://github.com/AgnostiqHQ/covalent/pull/726) some new metadata: `deps`, `call_before`, and `call_after`.  The Lepton constructor should be updated to add these metadata. In addition, `execution.py` could probably [handle](https://github.com/AgnostiqHQ/covalent/pull/726) missing metadata fields more gracefully.

PR #796, opened 2022-07-07
File transfer support for leptons

Emmanuel289:


AlejandroEsquivel:
Looking good @Emmanuel289 , you might need the changes from https://github.com/AgnostiqHQ/covalent/pull/742 to make it work however

Emmanuel289:
> Looking good @Emmanuel289 , you might need the changes from #742 to make it work however

Thanks @AlejandroEsquivel. Do I merge #742 directly into the branch or pull the changes after it is merged into develop?

tests/covalent_tests/workflow/lepton_test.py
```diff
@@ -0,0 +1,94 @@
...
+                    ],
+                )
+
+    assert mock_c_lepton.language == "C"
+    assert mock_c_lepton.library_name == temp_c_source_file.name
+    assert mock_c_lepton.function_name == "test_entry"
+    assert mock_c_lepton.argtypes == [("c_int", 0), ("LP_c_int", 2), ("LP_c_int", 1)]
+
+
+def test_lepton_languages(
```
venkatBala, line 71: why do we have the CamelCase and all lower case names for the languages. Is Julia/C++/Fortran also supposed to be on the list?

AlejandroEsquivel:
> > Looking good @Emmanuel289 , you might need the changes from #742 to make it work however
> 
> Thanks @AlejandroEsquivel. Do I merge #742 directly into this branch or pull the changes after it is merged into develop?

@Emmanuel289 You can #741 directly into this branch

Emmanuel289:
> 



> 

@AlejandroEsquivel sounds good, thanks

Emmanuel289:
pre-commit.ci autofix

AlejandroEsquivel (approved):

tests/covalent_tests/workflow/lepton_test.py
AlejandroEsquivel: These values are hardcoded in `_LANG_PY` and `_LANG_C` attributes of the Lepton class. They seem to be constants that determine if the supplied `language` kwarg to Lepton class matches their values and effectively makes up the supported languages.

If Julia/C++/Fortran are to be supported then I believe we need more constants in the Lepton class that reference these language but a separate issue for this should be created if that is the case.

PR #798, opened 2022-07-07
safe development db path

scottwn:


scottwn:
@FyzHsn this creates the DB on the `_dev` path, but no records are added to the DB. Should I expect the records to get added by `persist()` at this point?

covalent_dispatcher/_cli/service.py
```diff
@@ -304,6 +304,7 @@ def start(
         {
             "user_interface.address": "0.0.0.0",
             "user_interface.port": port,
+            "develop": develop,
```
scottwn, line 307: @FyzHsn @wjcunningham7 This one line is breaking all the tests. I can not figure out why. Any idea how to set this flag correctly?

FyzHsn: Hmmm It's also breaking the cli tools! Lemme see how we can get around it

FyzHsn: Yeah I know why this is breaking - there was a different reason to include the develop flag for CLI. Unfortunately, this is creating a clash. I will work on getting this working

covalent_dispatcher/_db/dispatchdb.py
```diff
@@ -167,6 +171,23 @@ def __exit__(self, exc_type, exc_value, traceback):
 
         return False
 
+    def _db_dev_path(self):
+        if get_config("develop"):
+            sqlite = ".sqlite"
+            path = "sqlite+pysqlite:///" + self._dbpath.split(sqlite)[0] + "_dev" + sqlite
+            print("***SET DB PATH FOR DEVELOPMENT***")
```
venkatBala, line 178: are these print statements included for debugging? could we use logger?

scottwn:
I did put those in stdout on purpose so it tells you where it's putting the
DB in develop mode but I'm fine with putting them in logs if you think it's
better

covalent_dispatcher/_service/app.py
```diff
@@ -77,3 +79,9 @@ def cancel() -> Response:
     dispatcher.cancel_running_dispatch(dispatch_id)
 
     return jsonify(f"Dispatch {dispatch_id} cancelled.")
+
+
+@bp.route("/db-path", methods=["GET"])
```
venkatBala, line 84: route endpoint `/db/path` to match the REST resource style better?

requirements.txt
```diff
@@ -1,6 +1,6 @@
 cloudpickle==2.0.0
 dask[distributed]==2022.6.0
-flask==2.0.1
+flask==2.1.2
```
venkatBala, line 3: was this version bump inadvertent?

covalent_dispatcher/_service/app.py
FyzHsn: We can come back to this as an enhancement later onwards!

requirements.txt
venkatBala: resolving for now... to be discussed during enhancements @scottwn

venkatBala (dismissed):

PR #799, opened 2022-07-07
DB Migrations Setup with Alembic & DataStore Updates

AlejandroEsquivel:

# Notes

Temporarily removed processing uploads, deletes during db session but can be added back when we tackle this in the future (see the [commit](https://github.com/AgnostiqHQ/covalent/pull/799/commits/f810d93c02b9943bc3d63b95526be0d3e628e431)).

## Using Workflow DB Session

```python
from sqlalchemy import select
from covalent._data_store.datastore import workflow_db
import covalent._data_store.models

with workflow_db.session() as session:
    statement = select(models.Lattice)
    all_lattices = session.execute(statement).all()
```

## Running migrations

To run migrations which have not yet run
```bash
alembic upgrade head
```

To view history of migrations

```bash
alembic history
```

[See more documentation around migrations](https://github.com/AgnostiqHQ/covalent/blob/8ccf322fbec0c597cd038c19a88f8bedd7b54f37/alembic/README.md)


covalent/_data_store/datastore.py
```diff
@@ -34,15 +34,15 @@
...
+        db_URL: Optional[str] = None,
         storage_backend_map: Dict[str, StorageBackend] = default_backend_map,
         initialize_db: bool = False,
         **kwargs,
     ):
         if db_URL:
             self.db_URL = db_URL
         else:
-            self.db_URL = "sqlite+pysqlite://" + get_config("workflow_data.db_path")
+            self.db_URL = "sqlite+pysqlite:///" + get_config("workflow_data.db_path")
```
AlejandroEsquivel, line 45: @scottwn @FyzHsn  It seemed that there was a missing slash in order to use an absolute path like the one specified by the config `workflow_data` (Refer to https://docs.sqlalchemy.org/en/14/core/engines.html#sqlite). Was this working for y'all before?

FyzHsn: No sir! It wasn't working. So far, we were manually inserting the DB URL. Thanks for adding the fix.

AlejandroEsquivel: Got it, thanks !

scottwn:
how often do the `versions` get committed to source? is there a strategy for this?

covalent_dispatcher/_db/dispatchdb.py
```diff
@@ -183,7 +183,6 @@ def __exit__(self, exc_type, exc_value, traceback):
 
     def _db_dev_path(self):
```
venkatBala, line 184: Based on recent changes made by @FyzHsn .. this may no longer be needed... do check with him if possible

venkatBala (dismissed):
looks good... but would just check with @FyzHsn to ensure if the `_db_dev_path` function is still needed!

FyzHsn (approved):
LGTM!

covalent_dispatcher/_db/dispatchdb.py
AlejandroEsquivel: got it, thanks @venkatBala , I removed this when fixing merge conflict

covalent/_data_store/models.py
```diff
@@ -179,7 +179,7 @@ class ElectronDependency(Base):
     arg_index = Column(Integer, nullable=True)
 
     # Name of the column which signifies soft deletion of the electron dependencies corresponding to a lattice
-    is_active = Column(Boolean, nullable=False)
+    is_active = Column(Boolean, nullable=False, default=True)
```
AlejandroEsquivel, line 182: @FyzHsn @scottwn wanted to check if these defaults for `is_active` are sensible

covalent/_data_store/models.py
```diff
@@ -76,11 +76,11 @@ class Lattice(Base):
     transport_graph_filename = Column(Text)
 
     # Name of the column which signifies soft deletion of a lattice
-    is_active = Column(Boolean, nullable=False)
+    is_active = Column(Boolean, nullable=False, default=True)
 
     # Timestamps
-    created_at = Column(DateTime, nullable=False)
-    updated_at = Column(DateTime, nullable=False)
+    created_at = Column(DateTime, nullable=False, server_default=func.now())
```
AlejandroEsquivel, line 82: defaults as per: https://docs.sqlalchemy.org/en/14/core/defaults.html#server-invoked-ddl-explicit-default-expressions

FyzHsn (approved):

covalent/_data_store/models.py
FyzHsn: Yeah, only when the user deletes a workflow in the UI can the field be changed to False

Issue #780, opened 2022-07-05
Support ct.electrons.bash


scottwn:
Requirements documented in Notion https://www.notion.so/Bash-support-tighter-610c5fd1418941d8879a1bb19158a089

PR #800, opened 2022-07-10
re-adding bash lepton support

wjcunningham7:


wjcunningham7:
Summary of modifications:
- Leptons which use scripting languages can be specified as either (1) a command run in the shell/console or (2) a call to a function in a library/script. Leptons which use compiled languages must specify a library and a function name.  In the future, compiled languages will also allow the user to specify compilation instructions via `CallDeps`.
- The keyword argument `display_name` can be used to override the name appearing in the UI. Particularly useful when the lepton is a command.
- All arguments except for `language` are now keyword arguments.
- Keyword arguments passed to a Bash Lepton are understood to define environment variables within the shell.
- Non-keyword arguments fill in $1, $2, etc.
- Named outputs enumerate variables within the shell which will be returned to the user. These can be either `Lepton.OUTPUT` or `Lepton.INPUT_OUTPUT` types.

wjcunningham7:
note this does not complete #780 entirely; there is an extra decorator needed which will be included in a separate PR

covalent/_workflow/lepton.py
```diff
@@ -87,6 +94,31 @@ def __init__(
         self.function_name = function_name
         # Types must be stored as strings, since not all type objects can be pickled
         self.argtypes = [(arg[0].__name__, arg[1]) for arg in argtypes]
+        print(self.argtypes)
```
scottwn, line 97: is this print statement intended?

scottwn (approved):

PR #823, opened 2022-07-13
removing the old dispatch DB

scottwn:

### DIFF
--- a/tests/workflow_stack_test.py
+++ b/tests/workflow_stack_test.py
@@ -131,7 +131,7 @@ def test_sublatticing(db):
     assert workflow_result.error is None
     assert workflow_result.status == str(Result.COMPLETED)
     assert workflow_result.result == 3
-    assert workflow_result.get_node_result(0)["sublattice_result"].result == 3
+    assert workflow_result.get_node_result(db, 0)["sublattice_result"].result == 3
 
 
 def test_internal_sublattice_dispatch():
@@ -642,7 +642,7 @@ def test_two_iterations_float():
     assert [0, 1, 2, 3, 4, 5, 6] == list(add_half_quarter.transport_graph._graph.nodes)
 
 
-def test_wait_for():
+def test_wait_for(db):
     """Test whether wait_for functionality executes as expected"""
 
     @ct.electron
@@ -672,21 +672,22 @@ def test_wait_for():
 
     dispatch_id = ct.dispatch(workflow)()
     result = ct.get_result(dispatch_id, wait=True)
-    rm._delete_result(dispatch_id)
+    result.persist(db)
 
     assert result.status == str(Result.COMPLETED)
     assert (
-        result.get_node_result(node_id=6)["start_time"]
-        > result.get_node_result(node_id=0)["end_time"]
+        result.get_node_result(db=db, node_id=6)["start_time"]
+        > result.get_node_result(db=db, node_id=0)["end_time"]
     )
     assert (
-        result.get_node_result(node_id=6)["start_time"]
-        > result.get_node_result(node_id=2)["end_time"]
+        result.get_node_result(db=db, node_id=6)["start_time"]
+        > result.get_node_result(db=db, node_id=2)["end_time"]
     )
     assert result.result == 1500
+    rm._delete_result(dispatch_id)
 
 
-def test_electron_getitem():
+def test_electron_getitem(db):
     """Test electron __getitem__, both with raw keys and with electron keys"""
 
     @ct.electron


scottwn:
note for future work, fix how the result status is compared. instead of casting things to strings, either override the compare for that type or make it a string natively

scottwn:
more future work: some text files are being pickled for convenience without `.pkl` extensions. this is confusing and should be changed.

scottwn:
_future work!_
- Improve error handling so exceptions are logged. There are many situations where exceptions are being dumped to null. "This is a deep structural issue" - @FyzHsn . There are existing issues for this that I will look up later.
- Log messages need to be timestamped, using standard logging format

covalent/_results_manager/result.py
```diff
@@ -282,13 +304,27 @@ def get_all_node_outputs(self) -> dict:
...
+                session.query(models.Lattice)
+                .where(models.Lattice.dispatch_id == self.dispatch_id)
+                .first()
+                .id
+            )
+            electron_records = (
+                session.query(models.Electron)
+                .where(models.Electron.parent_lattice_id == lattice_id)
+                .all()
+            )
```
scottwn, line 319: potential tuple issue

covalent/_results_manager/result.py
```diff
@@ -298,36 +334,117 @@ def get_all_node_results(self) -> List[Dict]:
...
+                session.query(models.Electron)
+                .where(
+                    and_(
+                        models.Electron.parent_lattice_id == lattice_id,
+                        models.Electron.transport_graph_node_id == node_id,
+                    )
+                )
+                .first()
+                .name
+            )
```
scottwn, line 383: potential tuple issue

covalent/_results_manager/result.py
```diff
@@ -298,36 +334,117 @@ def get_all_node_results(self) -> List[Dict]:
...
+            return (
+                session.query(models.Electron.status)
+                .where(
+                    and_(
+                        models.Electron.parent_lattice_id == lattice_id,
+                        models.Electron.transport_graph_node_id == node_id,
+                    )
+                )
+                .first()
+            )
```
scottwn, line 411: potential tuple issue

covalent/_results_manager/result.py
```diff
@@ -298,36 +334,117 @@ def get_all_node_results(self) -> List[Dict]:
...
+                .where(
+                    (
+                        and_(
+                            models.Electron.parent_lattice_id == lattice_id,
+                            models.Electron.transport_graph_node_id == node_id,
+                        )
+                    )
+                )
+                .first()
+            )
```
scottwn, line 443: potential tuple issue

covalent/_results_manager/result.py
```diff
@@ -352,11 +489,33 @@ def _get_node_error(self, node_id: int) -> Any:
...
```
scottwn, line 511: potential tuple issue

covalent/_results_manager/result.py
```diff
@@ -410,12 +582,36 @@ def _update_node(
             self.lattice.transport_graph.set_node_value(
                 node_id, "sublattice_result", sublattice_result
             )
+            # TODO (DBWORK) - Write sublattice record - not sure what to do here - assuming that this data gets written during the sublattice dispatch sync
```
scottwn, line 585: is this todo done?

covalent_dispatcher/_core/execution.py
```diff
@@ -470,20 +563,56 @@ def run_workflow(dispatch_id: str, results_dir: str, tasks_pool: ThreadPoolExecu
         None
     """
 
-    result_object = rm._get_result_from_file(dispatch_id, results_dir)
+    # TODO (DBWORK) - Needs the entire workflow - I do have to rehydrate the result object
```
scottwn, line 566: I think this is done

covalent_dispatcher/_core/execution.py
```diff
@@ -470,20 +563,56 @@ def run_workflow(dispatch_id: str, results_dir: str, tasks_pool: ThreadPoolExecu
...
+    # kwargs = result_object.inputs["kwargs"]
+    # app_log.warning(f"args: {args}, kwargs: {kwargs}")
+    # lattice.build_graph(*args, **kwargs)
+    # lattice.transport_graph = lattice.transport_graph.serialize()
+    # result_object = Result(lattice, lattice.metadata["results_dir"], dispatch_id)
+    # transport_graph = _TransportGraph()
+    # transport_graph.deserialize(result_object.lattice.transport_graph)
+    # result_object._lattice.transport_graph = transport_graph
+    # result_object._initialize_nodes()
+    # result_object.save()
```
scottwn, line 581: I think this is done

covalent/_results_manager/result.py
FyzHsn: Turns out (1) is 1 and (1,) will be a tuple

covalent/_results_manager/result.py
FyzHsn: I'll double check and remove it

covalent/_results_manager/results_manager.py
```diff
@@ -49,59 +57,105 @@ def get_result(
...
+                str(Result.CANCELLED),
+                str(Result.POSTPROCESSING_FAILED),
+                str(Result.PENDING_POSTPROCESSING),
             ]:
-                return result
+                return lattice_record.status if status_only else result_from(lattice_record)
+
         except (FileNotFoundError, EOFError, _pickle.UnpicklingError):
             if wait:
                 continue
```
scottwn, line 161: I think we should start here to figure out why everything is failing again

cjao: The ["Hello World" workflow](https://covalent.readthedocs.io/en/latest/getting_started/index.html#hello-covalent) now fails.

One problem is that `Lattice.function` is now [serialized when the workflow is first constructed](https://github.com/AgnostiqHQ/covalent/blob/88a56179b45f7400496a09510bd2530d8a23906b/covalent/_workflow/lattice.py#L100) and never deserialized by the server (we should probably add tests to ensure that the Covalent server never calls `TransportableObject.get_deserialized()`). Thus [this line](https://github.com/AgnostiqHQ/covalent/blob/88a56179b45f7400496a09510bd2530d8a23906b/covalent/_results_manager/results_manager.py#L102) no longer works because the retrieved `function` is a `TransportableObject`. The ordinary `Lattice` constructor tries to extract the function name by calling `self.function.__name__`. 

Suggestion: Don't directly invoke the Lattice constructor (which is expecting an ordinary function). After retrieving all the attributes from a `lattice_record`, create an attribute dictionary, create an dummy Lattice (e.g. with the identity function), and set its attributes via `lattice.__dict__ = [attribute dictionary]`

covalent_dispatcher/_core/execution.py
```diff
@@ -582,47 +631,58 @@ def task_callback(future: Future):
...
-                result_webhook.send_update(result_object)
-                return result_object
-
-    # post process the lattice
+        app_log.warning("8: Run_task finished running (run_planned_workflow)")
+
+        try:
+            # When one or more nodes failed in the last iteration, don't iterate further
+            for node_id in nodes:
+                if result_object._get_node_status(DispatchDB()._get_data_store(), node_id) == str(
```
cjao, line 639: Is the dispatcher server no longer going to have the result object in memory during workflow execution?

covalent/_results_manager/results_manager.py
FyzHsn: This is now fixed! Thank you @cjao for your suggestion. The next issue is related to postprocessing the sublattice and lattice results in my sample workflow. I imagine it has something to do with the code expecting a TransportableObject but getting a scalar value.

covalent/_results_manager/results_manager.py
```diff
@@ -49,59 +59,114 @@ def get_result(
...
+        transport_graph = pickle.loads(f.read())
+    with open(
+        os.path.join(lattice_record.storage_path, lattice_record.results_filename), "rb"
+    ) as f:
+        output = pickle.loads(f.read())
+
+    attributes = {
+        "metadata": {"executor": executor},
+        "args": inputs["args"],
+        "kwargs": inputs["kwargs"],
```
cjao, line 107: fyi, the lattice object now has two [additional attributes](https://github.com/AgnostiqHQ/covalent/blob/378131b777de2185c70c52a483cdfa54c3952875/covalent/_workflow/lattice.py#L94-L95): `named_args` and `named_kwargs`. These were previously computed by  `result_webhook.py` with each node update from the lattiice args and kwargs, but doing so requires access to the unserialized workflow inputs. Hence they are now [stored](https://github.com/AgnostiqHQ/covalent/commit/f06f84a0cc487a85ac88177c903f898fe317f845#diff-bfbac3709c395f2f264b85d0a49527ee0e1f11a52a1e2484d28f4fc867b82744R231) in the lattice during build_graph and simply [read by `result_webhook.py`](https://github.com/AgnostiqHQ/covalent/commit/f06f84a0cc487a85ac88177c903f898fe317f845#diff-6a95a7b8ccaf3877a43ca2eb8debac479d7cfecd82a12699b82d8239511137c7L86)

cjao: [This](https://github.com/AgnostiqHQ/covalent/blob/378131b777de2185c70c52a483cdfa54c3952875/covalent/_results_manager/result.py#L728) is the problematic line. `Result.result` tries to deserialize the result and should never be called by the server. Instead use `Result.encoded_result`.

covalent/_results_manager/results_manager.py
cjao: [This](https://github.com/AgnostiqHQ/covalent/blob/378131b777de2185c70c52a483cdfa54c3952875/covalent/_results_manager/result.py#L728) is the problematic line. `Result.result` tries to deserialize the result and should never be called by the server. Instead use `Result.encoded_result`.

covalent_dispatcher/_core/execution.py
FyzHsn: We'll still keep the result object in ram. Some of these statements are unnecessary and we need to take these out, but I need to merge this branch ASAP.

covalent/_results_manager/results_manager.py
FyzHsn: Yessir! These are changed. Still getting some related errors tho
Screen Shot 2022-07-17 at 2 16 50 PM
:

cjao: Can you confirm that the node outputs in the transport graph are TransportableObjects before starting post-processing?

covalent_dispatcher/_core/execution.py
```diff
@@ -668,25 +726,38 @@ def task_callback(future: Future):
...
         return result_object
 
     pp_end_time = post_process_result["end_time"]
     app_log.debug(f"Post-processing completed at {pp_end_time}")
-
     result_object._result = post_process_result["output"]
 
-    # app_log.debug(f"Result: {result_object._result}")
-
+    try:
```
cjao, line 738: Hi @FyzHsn , this block shouldn't be here. This was the old postprocessing step which is now [run in an executor](https://github.com/AgnostiqHQ/covalent/commit/f06f84a0cc487a85ac88177c903f898fe317f845#diff-a99ca6e887d5a6439efcda540e9e2661cdafa9e3a92109f2ac9c390299f794fbL424).

FyzHsn: Oooh thanks, took this out!

covalent/_results_manager/results_manager.py
FyzHsn: Are there any actionable items on my side or can it wait?

FyzHsn:
@cjao What's the purpose of the `post_process` method added to the Result object?

covalent/_results_manager/results_manager.py
cjao: Might be an issue for the UI which pulls the result object from DB - [`encode_result`](https://github.com/AgnostiqHQ/covalent/issues/748) won't work. But I'd say that if tests pass, this can be deferred to v10 or v11 of the schema -- whenever the UI needs to work.

cjao:
> @cjao What's the purpose of the `post_process` method added to the Result object?

It allows the user to post-process the workflow client-side (i.e. without the Covalent server running). The changes to postprocessing are discussed [here](https://github.com/AgnostiqHQ/covalent/issues/748#issuecomment-1172934487):


>* Postprocessing, which involves deserializing and executing the workflow function on the deserialized workflow inputs, becomes a task to be run in a designated executor whose environment satisfies the workflow dependencies. Lattices now expose a workflow_executor property analogous to the existing executor property. In addition to the usual executors, workflow_executor supports a special value called "client". Setting workflow_executor="client" causes the dispatcher to halt workflow execution just before preprocessing and set the workflow status to Pending Postprocessing. The client can take the Result, which contains only encoded node outputs, and postprocess the workflow client-side by calling Result.post_process(). This returns the deserialized workflow result.
>
>* Since postprocessing is now decoupled from the main workflow execution, the Result has acquired some new statuses: POSTPROCESSING, PENDING_POSTPROCESSING, FAILED_POSTPROCESSING. The last one would be encountered if the workflow_executor specified by the user doesn't actually satisfy the workflow dependencies, or if the postprocessing fails for any other reason; for instance, a remote executor might time out if the workflow contains a long sleep() statement. Whatever the reason, one can still reattempt postprocessing client-side.

venkatBala (approved):

FyzHsn (dismissed):

FyzHsn:
Also for future work: 

- [ ] Ensure that some of the database fields are not stored as TransportableObjects() but rather converted to after being read from storage.
- [ ] Remove app_log.warning statements OR convert them to app_log.debug statements.

venkatBala (approved):

PR #844, opened 2022-07-15
File Transfer Keyword Args

AlejandroEsquivel:


wjcunningham7:
This pull request is stale because it has been in review for at least three days. It will be closed in two more days.

wjcunningham7:
This pull request is stale because it has been in review for at least three days. It will be closed in two more days.

covalent/_workflow/lepton.py
```diff
@@ -285,6 +292,11 @@ def shell_wrapper(*args, **kwargs) -> Any:
             import builtins
             import subprocess
 
+            # remove reserved retval kwargs from kwargs in lepton (ex. 'files')
```
cjao, line 295: If `files` is considered an illegal parameter name, wouldn't it be better to fail loudly here?

covalent/executor/base.py
```diff
@@ -65,9 +67,20 @@ def wrapper_fn(
         cb_args = serialized_args.get_deserialized()
         cb_kwargs = serialized_kwargs.get_deserialized()
         retval = cb_fn(*cb_args, **cb_kwargs)
-        if retval_key:
+        if retval_key and retval_key in cb_retvals:
+            if isinstance(cb_retvals[retval_key], list):
+                cb_retvals[retval_key].append(retval)
+            else:
+                cb_retvals[retval_key] = [cb_retvals[retval_key], retval]
```
cjao, line 74: Are we assuming here that no other call_before will have a return value key of `files`? If so, this seems ripe for unexpected behavior; do we know that `cb_retvals[retval_key]` is in fact a path intended for file transfer. If the `files` key is to be used exclusively by File Transfer, we should enforce that at workflow construction.

covalent/_workflow/depscall.py
```diff
@@ -23,6 +23,8 @@
 from .deps import Deps
 from .transport import TransportableObject
 
+RESERVED_RETVAL_KEY__FILES = "files"
```
cjao, line 26: I wonder if a less generic name would be appropriate here, like `file_transfer_files`. Electrons might implement file processing logic in addition to that provided by file transfer.

covalent/_workflow/electron.py
```diff
@@ -569,11 +569,17 @@ def electron(
...
+        _file_transfer_pre_hook_, _file_transfer_call_dep_ = file_transfer.cp()
+
+        # pre-file transfer hook to create any necessary temporary files
+        internal_call_before_deps.append(
+            DepsCall(_file_transfer_pre_hook_, retval_keyword=RESERVED_RETVAL_KEY__FILES)
+        )
+
         if file_transfer.order == Order.AFTER:
-            internal_call_after_deps.append(DepsCall(_callback_))
+            internal_call_after_deps.append(DepsCall(_file_transfer_call_dep_))
```
cjao, line 580: See my comment on `base.py` about adding a check here that no other call_before uses `RESERVED_RETVAL_KEY__FILES` as a return value key.

cjao (changes requested):

covalent/_workflow/depscall.py
AlejandroEsquivel: The `files` keyword was chosen to be concise as users will have to use this kwarg everywhere there is a file transfer to get the abstracted filepaths I think making it longer would degrade user experience. Could you maybe elaborate on what type of file processing logic would occur so we can weigh the importance and if this would also be a user facing feature?
cheers @cjao

covalent/executor/base.py
AlejandroEsquivel: Good points @cjao, currently if there are non-unique retval keys the intention is to accumulate them in an array even if they do not correspond to reserved retval keys (such as `files`) - which would be included in RTD, however I think the way the logic is written would lead to some unexpected behavior if we user's return values are arrays I will revise this and cover it with tests.

I agree that during workflow construction we should check if the user is supplying reserved retval keys and fail hard during workflow construction - will include this.

covalent/_workflow/electron.py
AlejandroEsquivel: Noted @cjao

covalent/_workflow/depscall.py
cjao: This is mostly a hypothetical concern and we can revisit it down the road if it does become a problem.

covalent/_workflow/lepton.py
AlejandroEsquivel: This was a temporary work around which is now removed @cjao , it seems that `DepsCall` return values are not supported in general by Leptons. Since this is out of scope for this PR there will be a separate issue to address this.

For now I am throwing an exception when Leptons are created with `DepsCall` objects that have a truthy `retval_keyword` attribute.

cjao (approved):
Looks good @AlejandroEsquivel !

PR #861, opened 2022-07-18
Enhance get_result to use HTTP

scottwn:


covalent/_results_manager/results_manager.py
```diff
@@ -133,50 +133,38 @@ def dummy_function(x):
...
-            ]:
-                return lattice_record.status if status_only else result_from(lattice_record)
-
-        except (FileNotFoundError, EOFError, _pickle.UnpicklingError):
-            if wait:
-                continue
-            raise RuntimeError(
-                "Result not ready to read yet. Please wait for a couple of seconds."
-            )
+    adapter = HTTPAdapter()
```
venkatBala, line 158: should add a `max_retries=<something>` here to automate retries?

covalent_dispatcher/_service/app.py
```diff
@@ -85,3 +91,63 @@ def cancel() -> Response:
...
+                output = {
+                    "id": dispatch_id,
+                    "status": lattice_record.status,
+                }
+                if not status_only:
+                    output["result"] = codecs.encode(
+                        pickle.dumps(result_from(lattice_record)), "base64"
+                    ).decode()
+                return jsonify(output)
+            elif lattice_record.status in [
```
venkatBala, line 124: can the `elif not wait` and `elif lattice_record.status` be combined to a single `else`... seems like we are performing the exact same action in the body in both cases?

covalent/_results_manager/results_manager.py
scottwn: I think retries are enabled automatically, even without `max_retries`?

venkatBala (changes requested):

FyzHsn (approved):

covalent/_results_manager/results_manager.py
FyzHsn: Yes, I think retries are automatic. To set more custom limits on retry: https://stackoverflow.com/questions/15431044/can-i-set-max-retries-for-requests-request

FyzHsn: I recommend opening another PR if we need to make enhancements. Resolving and merging this PR ASAP.

FyzHsn (approved):

venkatBala (approved):

mshkanth:
Hi @scottwn - We noticed that have added a server end point to re-construct the result object from the new DB. This right now uses the Flask blue print that is to be removed/de-commissioned given that we are re-factoring everything to Fast API (https://github.com/AgnostiqHQ/covalent/discussions/852). We will go ahead and port this end point also to FastAPI.

Cc @cjao @FyzHsn

FyzHsn:
> Hi @scottwn - We noticed that have added a server end point to re-construct the result object from the new DB. This right now uses the Flask blue print that is to be removed/de-commissioned given that we are re-factoring everything to Fast API (#852). We will go ahead and port this end point also to FastAPI.
> 
> Cc @cjao @FyzHsn

Excellent @mshkanth! That works for us!

mshkanth:
> > Hi @scottwn - We noticed that have added a server end point to re-construct the result object from the new DB. This right now uses the Flask blue print that is to be removed/de-commissioned given that we are re-factoring everything to Fast API (#852). We will go ahead and port this end point also to FastAPI.
> > Cc @cjao @FyzHsn
> 
> Excellent @mshkanth! That works for us!

@ArunPsiog @mpvgithub @Aravind-psiog - Pls include this in your re-factoring

PR #883, opened 2022-07-20
Simplify executor UX

### LINKED ISSUE #771: Simplification of Executor UX
## Acceptance Criteria

- As a covalent user I should be able to create custom executor plugins with minimal boilerplate code

## Notes

Currently there is a lot of boiler plate for executer plugin, We need to make it easy to create custom executers without all the fancy bells. For example, all the user needs to construct could be made into this

```jsx
class BaseAsyncExecuter:
	...

something = executer.pull()
```

## User

```python
class MyExecuter(BaseAsyncExecuter):
	def __init__(cores):
			self.cores=cores
			
	
	def execute(self,func,*args,**kwargs)
			print(self.cores)
      result=func(*args,**kwargs)
			
			result+=self.executer_params
      return result

	def pull(self):
    ....
			

executer=MyExecuter(cores=3)
```

In baseexecuter, we can wrap this self.execute function with something else that does various things we are doing right now, like setting status to running, changing conda env, module loading etc... Ideally we want users to do as minimal things as possible to create plugins. Inspiration from prefect’s

cjao:

Closes #771


cjao:
Something funny is happening with codecov. The "Impacted Files" list appears to be the entire repo.

scottwn:
I see that, very weird. The UI change renders Codecov essentially useless. I think you'll need to open a support ticket with Codecov.
The Codecov annotations on this PR appear to be accurate tho and you can use those to increase the patch coverage.

venkatBala (approved):
looks good to me... the UX has been simplified greatly

venkatBala (approved):

PR #885, opened 2022-07-20
Read node details from transport graph instead of DB

FyzHsn:


scottwn:
What's the motivation for this? Shouldn't we be relying on the DB for node data?

covalent/_results_manager/result.py
```diff
@@ -303,37 +299,22 @@ def get_node_result(self, db: DataStore, node_id: int) -> dict:
...
-        #         node_id, "sublattice_result"
-        #     ),
-        #     "stdout": self.lattice.transport_graph.get_node_value(node_id, "stdout"),
-        #     "stderr": self.lattice.transport_graph.get_node_value(node_id, "stderr"),
-        # }
-
         return {
             "node_id": node_id,
-            "node_name": self._get_node_name(db, node_id),
+            "node_name": self.lattice.transport_graph.get_node_value(node_id, "node_name"),
```
venkatBala, line 304: Will there be time when there can be a lookup failure i.e "node_name" not present, to cover that can we return None/something equivalent?
```suggestion
            "node_name": self.lattice.transport_graph.get_node_value(node_id, "node_name") or None
```

covalent/_results_manager/result.py
```diff
@@ -425,27 +374,9 @@ def _get_node_name(self, db: DataStore, node_id: int) -> str:
...
-                        models.Electron.transport_graph_node_id == node_id,
-                    )
-                )
-                .first()
-                .name
-            )
+        return self.lattice.transport_graph.get_node_value(node_id, "node_name")
 
-    def _get_node_status(self, db: DataStore, node_id: int) -> "Status":
+    def _get_node_status(self, node_id: int) -> "Status":
```
venkatBala, line 379:
```suggestion
    def _get_node_status(self, node_id: int) -> Status:
```

venkatBala (commented):
just a few questions/suggestions

PR #896, opened 2022-07-21
Second Attempt at Resolving 693

### LINKED ISSUE #693: Support async executors

kessler-frost:

Description imported from #694 

* Run `_run_planned_workflow` in an `asyncio` event loop.
* In `_run_task`, await on `executor.execute` (preferred) if `BaseAsyncExecutor` subclassed executor is used, else call it in a `ThreadPoolExecutor`.
* Sublattices are also dispatched as an `asyncio.Task` awaitable.
* Simplified `executor.execute` so that `call_before` and `call_after` are part of the wrapped callable.
* Make `DaskExecutor` async-aware.

Resolves #693.


covalent/executor/base.py
```diff
@@ -467,3 +462,48 @@ def get_conda_path(self) -> bool:
...
+        dispatch_id: str,
+        results_dir: str,
+        node_id: int = -1,
+    ) -> Any:
+        awaitable_run, out, err = super().execute(
+            function, args, kwargs, dispatch_id, results_dir, node_id
+        )
+
+        if not asyncio.iscoroutine(awaitable_run):
+            return (awaitable_run, out, err)
```
cjao, line 482: What do you think of making  `BaseAsyncExecutor` its own base class instead of a subclass of `BaseExecutor`? That way we can enforce that `run()` is a coroutine. Trying to handle the case where `run()` is not a coroutine here would block the event loop.

cjao: We can make a simple `BaseAsyncExecutor` base class with just `execute` and `run`. All the logic pertaining to a local conda env can be jettisoned as it is incompatible with async.

cjao (changes requested):

covalent/executor/base.py
cjao: Since an async `run()` is not really "derived" from non-async `run()` (they aren't invoked in the same way), `BaseAsyncExecutor` is logically parallel to, not derived from, `BaseExecutor`.

cjao:
Hi @kessler-frost, aside from my above reservation about deriving `BaseAsyncExecutor` from the synchronous `BaseExecutor`, everything else looks good!

covalent/executor/base.py
kessler-frost: >What do you think of making `BaseAsyncExecutor` its own base class instead of a subclass of `BaseExecutor`?

I would've agreed with that, actually I was going to do it like that at first, but a lot of things in `BaseAsyncExecutor` are the same as in `BaseExecutor` which are unrelated to `conda`, and their respective function signature, attribute types are similar as well; things like: `log_stdout`, `log_stderr`, `cache_dir`, `get_dispatch_context` (although this will be removed soon), `write_streams_to_file()`, `to_dict()`, `from_dict()`, `short_name()`.

Since all of these functions/attributes will need to be copy-pasted in the new `BaseAsyncExecutor`, I figured it was better to derive them from `BaseExecutor`. Perhaps if we put those things in a common file then both executors can import it from there? Might be out of the scope of this issue.

>That way we can enforce that run() is a coroutine. Trying to handle the case where run() is not a coroutine here would block the event loop.

Yeah, I agree that it will block the eventloop here, but only in the case when the `conda_env` is specified.
Will creating a separate class allow us to enforce `run()` being a coroutine in a way which cannot be done now?

>Since an async `run()` is not really "derived" from non-async `run()` (they aren't invoked in the same way), `BaseAsyncExecutor` is logically parallel to, not derived from, `BaseExecutor`.

Agreed, but it's actually the other stuff that is derived from `BaseExecutor` not `run()`.

kessler-frost: I also like the idea of replacing how conda related stuff is being handled right now in executors. I think we can improve it.

cjao: >Since all of these functions/attributes will need to be copy-pasted in the new BaseAsyncExecutor, I figured it was better to derive them from BaseExecutor. Perhaps if we put those things in a common file then both executors can import it from there? Might be out of the scope of this issue.

cjao: This restructuring can be left for a followup issue. I'm happy to approve this for now.

kessler-frost: >Maybe something like
`AbstractBaseExecutor` (common methods)
`BaseExecutor` `BaseAsyncExecutor`

Yeah, I like that.

cjao:
Documenting some above comments for followup:
>What do you think of making BaseAsyncExecutor its own base class instead of a subclass of BaseExecutor? That way we can enforce that run() is a coroutine. Trying to handle the case where run() is not a coroutine here would block the event loop.

>We can make a simple BaseAsyncExecutor base class with just execute and run. All the logic pertaining to a local conda env can be jettisoned as it is incompatible with async.


> Since an async run() is not really "derived" from non-async run() (they aren't invoked in the same way), BaseAsyncExecutor is logically parallel to, not derived from, BaseExecutor.

venkatBala (changes requested):

covalent/executor/executor_plugins/dask.py
```diff
@@ -87,12 +88,23 @@ def __init__(
 
         self.scheduler_address = scheduler_address
 
-    def run(self, function: callable, args: List, kwargs: Dict):
+    async def run(self, function: Callable, args: List, kwargs: Dict):
         """Submit the function and inputs to the dask cluster"""
-        dask_client = get_client(address=self.scheduler_address, timeout=1)
+
+        dask_client = _address_client_mapper.get(self.scheduler_address)
```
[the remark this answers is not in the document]
kessler-frost: So basically the `get_client` function of Dask is very weirdly broken, it's near impossible to get an asynchronous client from there consistently, which is discussed in a PR @cjao created https://github.com/dask/distributed/pull/6724 to allow that. But they suggested that what we are trying to do may be better achieved if we maintain our own global dictionary of clients.

Oh, I forgot to mention that it is not possible to create a large number of `Client` objects (whereas we can have a large number of electrons), so, now what happens is if we don't have a `Client` object already stored in that dictionary, we create a new asynchronous client and store it in that dictionary.

Now, you may ask why isn't it thread-safe, so I've tried doing that but to no success (using MPQs). Still, there haven't been any issues even when I ran multiple concurrent workflows.

covalent/_shared_files/utils.py
```diff
@@ -197,3 +197,6 @@ def get_named_params(func, args, kwargs):
                     named_kwargs[key] = value
 
     return (named_args, named_kwargs)
+
+
+_address_client_mapper = {}
```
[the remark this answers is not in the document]
kessler-frost: Added.

cjao (approved):

venkatBala (approved):

PR #897, opened 2022-07-22
add abstract setup/teardown executor hooks

venkatBala:


cjao (dismissed):

covalent/executor/base.py
```diff
@@ -28,6 +28,7 @@
 import subprocess
 import tempfile
 from abc import ABC, abstractmethod
+from ast import Call
```
cjao, line 31: What is this used for?

cjao (changes requested):

covalent/executor/base.py
venkatBala: ah my bad... bad typing import

cjao (approved):

covalent/executor/base.py
```diff
@@ -493,6 +510,14 @@ async def execute(
 
         return (result, stdout.getvalue(), stderr.getvalue())
 
+    async def setup(self):
```
cjao, line 513: Are you planning to integrate `setup` and `teardown` with `execute()`?

cjao (changes requested):

wjcunningham7:
This pull request is stale because it has been in review for at least three days. It will be closed in two more days.

covalent/executor/base.py
wjcunningham7: the dispatcher should call setup and teardown before/after execute, if they exist

cjao (approved):

PR #927, opened 2022-07-27
Create result in entrypoint

### LINKED ISSUE #926: Don't assume in SDK get_result that dispatch_id actually exists
The SDK's `get_result()` [retries indefinitely](https://github.com/AgnostiqHQ/covalent/blob/f50b979ca8bba90de9ca61e879cbc6589e8ae3d1/covalent/_results_manager/results_manager.py#L72-L78) if the `dispatch_id` is not found by the server. However, the specified `dispatch_id` might actually be invalid. We should just error out instead.

cjao:

Closes #926 since the server would only fail to find the `dispatch_id` if the dispatch doesn't actually exist (or has been deleted), not because the dispatch has been received but not yet saved to the db.


covalent/_results_manager/results_manager.py
```diff
@@ -69,12 +64,12 @@ def get_result(dispatch_id: str, wait: bool = False) -> Result:
         )
         result_object = pickle.loads(codecs.decode(result["result"].encode(), "base64"))
 
-    except MissingLatticeRecordError as e:
-        app_log.warning(
-            f"Dispatch ID {dispatch_id} was not found in the database. Either the Dispatch ID is incorrect or wait a couple of seconds before trying again."
+    except MissingLatticeRecordError:
+        app_log.debug(
+            f"Dispatch ID {dispatch_id} was not found in the database. Incorrect dispatch id."
```
scottwn, line 69: Assuming the `404` is returned correctly from the server then this actually _should_ be a warning.

covalent/_results_manager/results_manager.py
```diff
@@ -69,12 +64,12 @@ def get_result(dispatch_id: str, wait: bool = False) -> Result:
...
         )
 
-        time.sleep(0.01)
+        time.sleep(0.1)
         return get_result(dispatch_id, wait)
```
scottwn, line 73: Related to above... if the `404` is returned correctly, then I think we shouldn't be doing `get_result` again since it will keep `404`ing

covalent/_results_manager/results_manager.py
```diff
@@ -155,13 +150,12 @@ def _get_result_from_dispatcher(
     adapter = HTTPAdapter(max_retries=5)
     http = requests.Session()
     http.mount("http://", adapter)
-    url = "http://" + dispatcher + "/api/result/" + dispatch_id
```
scottwn, line 158: any particular reason you removed the `f` string?

covalent/_results_manager/results_manager.py
```diff
@@ -155,13 +150,12 @@ def _get_result_from_dispatcher(
...
-    response = http.get(url, params={"wait": wait, "status_only": status_only}, timeout=2)
+    url = f"http://{dispatcher}/api/result/{dispatch_id}"
+    response = http.get(url, params={"wait": wait, "status_only": status_only}, timeout=60)
```
scottwn, line 154: are you sure this is the right place for `timeout` and not `HTTPAdapter`?

scottwn (changes requested):

covalent/_results_manager/results_manager.py
cjao: Hi @scottwn , this is tracked in the followup issue #926. I'm rolling the fix for that into this branch.

covalent/_results_manager/results_manager.py
cjao: `timeout` is [a parameter](https://requests.readthedocs.io/en/latest/api/) of `Request`, not `HTTPAdapter`

covalent/_results_manager/results_manager.py
cjao: Fixed, thanks

covalent/_results_manager/results_manager.py
cjao: It's [still there](https://github.com/AgnostiqHQ/covalent/pull/927/files#diff-5694ed1aa0e5cf6e57cf19bb998f92485940c7df6209fa3b0d719a33fbb38802R152).

covalent/_results_manager/results_manager.py
cjao: I'm actually not sure if `timeout` is even appropriate here (what if a dispatch really does take a really long time?), but I think that's outside the scope of this issue.

FyzHsn (approved):

covalent/_results_manager/results_manager.py
```diff
@@ -155,13 +149,12 @@ def _get_result_from_dispatcher(
...
```
scottwn, line 153:
```suggestion
    response = http.get(url, params={"wait": wait, "status_only": status_only}, timeout=2)
```

covalent/_results_manager/results_manager.py
```diff
@@ -155,13 +149,12 @@ def _get_result_from_dispatcher(
...
     if response.status_code == 404:
         raise MissingLatticeRecordError
     response.raise_for_status()
-    result = response.json()
-    return result
+    return response.json()
```
scottwn, line 157:
```suggestion
    result = response.json()
    return result
```

FyzHsn (approved):

scottwn (approved):

PR #940, opened 2022-07-29
Fixing covalent pre-release

kessler-frost:


kessler-frost:
After going through `dorny/paths-filter`'s code I found out that they are themselves checking whether the current action being run is on a PR or not. If the action is being run on a PR then they use Github's API to detect file changes, otherwise they use a separate mechanism which involves doing a `git branch --show-current`.

So, there are two viable solutions that come to my mind:
1. We don't run this action if the thing that we are running on is not a PR, or
2. We install `buster-backports` and enable `git branch --show-current` on Buster.

I prefer option 1, since we have already run this action once in the PR, and the tests are still going to run even without it if I'm not misunderstanding. But @scottwn @wjcunningham7 It's up to you guys to decide.

wjcunningham7:
> After going through `dorny/paths-filter`'s code I found out that they are themselves checking whether the current action being run is on a PR or not. If the action is being run on a PR then they use Github's API to detect file changes, otherwise they use a separate mechanism which involves doing a `git branch --show-current`.
> 
> So, there are two viable solutions that come to my mind:
> 1. We don't run this action if the thing that we are running on is not a PR, or
> 2. We install `buster-backports` and enable `git branch --show-current` on Buster.
> 
> I prefer option 1, since we have already run this action once in the PR, and the tests are still going to run even without it if I'm not misunderstanding. But @scottwn @wjcunningham7 It's up to you guys to decide.

Let's just run all tests on develop. The filter was only needed to speed up PRs.

kessler-frost:
I see, so you're suggesting we remove this action altogether?

kessler-frost:
I think that's a good idea because even if we fix it for buster, it is still giving issues for centos. Removing this action in this PR.

wjcunningham7 (approved):

scottwn:
the suggestion is to only run this action if the trigger is `pull-request`, so it doesn't run on `workflow-run`

PR #949, opened 2022-08-02
enabling codecov carryforward

wjcunningham7:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -73,9 +73,11 @@ jobs:
             dispatcher:
             - 'covalent_dispatcher/**'
             ui_backend:
-            - 'covalent_ui/**'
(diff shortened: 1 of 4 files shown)


wjcunningham7:
This PR is an enhancement to `tests.yml` along with a fix to how Codecov is being used.

Summary of Changes:
- Codecov carryforward flags are enabled so that when certain tests aren't run on a PR, a previous report is carried forward.  Read more [here](https://docs.codecov.com/docs/carryforward-flags).  These changes are shown in `codecov.yml`.
- Front-end dependencies are only installed, and the front-end is only built, either when there have been changes to source, or when the whole test suite is being run on `develop`.
- The package validation step only runs on `develop`.

FyzHsn (approved):

Issue #35, opened 2022-01-28
Handle reading the result better in  get_result(wait=True).


wjcunningham7:
@santoshkumarradha commented on [Wed Dec 15 2021](https://github.com/AgnostiqHQ/covalent-staging/issues/214)

## Overview 

Currently we do https://github.com/AgnostiqHQ/covalent-staging/blob/9348253dbb3037616e18c12b7b5b0d83c3f21b62/covalent/_results_manager/results_manager.py#L68 to check if result is updated. This is bit of a bad design as when running in wait mode, we are reading a file that is actively be written to by the dispatcher without proper flushing. Ideally we would want to wait till the file has been modified and then read it again and flush it properly and not read it infinitely in a while loop. We should use more standard observer tools/patterns for watching like watchdog - https://github.com/gorakhargosh/watchdog

wjcunningham7:
@scottwn I believe option 2 will be minimal work with a reasonable improvement.  If so, please proceed with that option.

PR #959, opened 2022-08-02
removing while true in get_result

scottwn:


covalent_dispatcher/_service/app.py
```diff
@@ -97,46 +97,32 @@ def get_result(dispatch_id) -> Response:
...
+        if not status_only:
+            output["result"] = codecs.encode(
+                pickle.dumps(result_from(lattice_record)), "base64"
+            ).decode()
+        return jsonify(output)
+    response = make_response(
+        jsonify({"message": "Result not ready to read yet. Please wait for a couple of seconds."}),
+        503,
+    )
+    response.retry_after = 2
```
cjao, line 127: Does this effectively change the behavior of `get_result(wait=True)`? Since `ct.get_result()` [retries at most 5 times](https://github.com/AgnostiqHQ/covalent/blob/7da4139384fdadca3498790417ea027202d8863d/covalent/_results_manager/results_manager.py#L169), `dispatch_sync` would no longer work with workflows that last more than 10 seconds.

cjao (changes requested):
Does

covalent_dispatcher/_service/app.py
scottwn: The documentation for `max_retries` says

> Note, this applies only to failed DNS lookups, socket connections and connection timeouts, never to requests where data has made it to the server.

so I think we're good?

https://requests.readthedocs.io/en/latest/api/#requests.adapters.HTTPAdapter

cjao: Ah, I see. Thanks for the clarification

scottwn:
@cjao would it help if I wrote a test for the `wait` behavior? or do we already have a functional test for `dispatch_sync`?

cjao:
>@cjao would it help if I wrote a test for the wait behavior? or do we already have a functional test for dispatch_sync?

I actually just tried this:
```python
import covalent as ct
import time

@ct.electron
def task(x):
    print("sleeping for 20 s")
    time.sleep(20)
    return x

@ct.lattice
def workflow(x):
    return task(x)

res = ct.dispatch_sync(workflow)(5)
print(res.result)
```
This errors out after 10 seconds :(. It works if the sleep statement is changed to 5 seconds.

scottwn:
ok let me fix that, and also make a test for it. thanks @cjao

cjao:
Can we add custom retry logic in [results manager](https://github.com/AgnostiqHQ/covalent/blob/7da4139384fdadca3498790417ea027202d8863d/covalent/_results_manager/results_manager.py#L174) if a 503 status code is returned?

scottwn:
I think all it takes is reconfiguring the adapter. I'm trying to do it without introducing another `while` loop... since the point of #35 was to remove the while loop

scottwn:
this PR is currently on hold per @wjcunningham7

santoshkumarradha (approved):
LGTM from UX

covalent/_dispatcher_plugins/local.py
```diff
@@ -128,7 +129,7 @@ def wrapper(*args, **kwargs) -> Result:
 
             return get_result(
                 LocalDispatcher.dispatch(lattice, dispatcher_addr)(*args, **kwargs),
-                wait=True,
+                wait=wait.EXTREME,
```
cjao, line 132: This solution certainly works for all intents and purposes.

cjao (dismissed):

cjao (dismissed):

scottwn:
pre-commit.ci autofix

cjao (dismissed):

cjao (approved):

AlejandroEsquivel (approved):

PR #960, opened 2022-08-03
Resolve overwriting config values

### LINKED ISSUE #879: Covalent config is improperly overwritten
### Environment

- **Covalent version**: 0.142.0
- **Python version**: 3.8
- **Operating system**: Gentoo

### What is happening?

The dispatcher address is improperly overwritten when the server is started.  I'm not sure if it also happens to other values in the config.

1. Generate a default config file, e.g., using `covalent start`.
2. Stop the Covalent server with `covalent stop`
3. Manually edit the address in the config.
4. Start the server with `covalent start`.
5. Inspect the config to see "localhost" has overwritten the provided value.

### What should happen?

A user's config file should take precedence over any default values.  In addition, the server always seems to start on localhost, which is hard-coded into `covalent_ui/app.py`, even though it should really reference the config file.  E.g. in my case I want it to start on `0.0.0.0` instead of `localhost`.

### Any suggestions?

_No response_

FyzHsn:
Fixes #879 


wjcunningham7:
1. The call to `set_config` in `start` should not override the dispatcher or UI addresses
2. The same call should not set the port(s) where it's currently happening. The port should be set after the call to `next_available_port` in `graceful_start`
3. The dask options are properly set, but should only be set if `no_cluster == false`
4. The `no_cluster` variable is currently typed as a string but should be a bool
5. The variable `no_cluster_flag` seems redundant and can be removed

tests/covalent_dispatcher_tests/_cli/service_test.py
```diff
@@ -220,7 +220,7 @@ def test_start(mocker, monkeypatch, is_migration_pending, ignore_migrations):
 
     if ignore_migrations or not is_migration_pending:
         graceful_start_mock.assert_called_once()
-        set_config_mock.assert_called_once()
+        set_config_mock.assert_called()
```
venkatBala, line 223:
```suggestion
        assert set_config.call_count == 2
```
this might be a more stricter check to assert the exact number of times set_config is called within the `service.py`

venkatBala (changes requested):
just a small comment regarding assertion of the calling of the `set_config` mock

covalent_dispatcher/_cli/service.py
```diff
@@ -174,13 +174,9 @@ def _graceful_start(
...
     dev_mode_flag = "--develop" if develop else ""
-    no_cluster_flag = "--no-cluster"
     port = _next_available_port(port)
-    if no_cluster_flag in sys.argv:
-        launch_str = f"{pypath} python app.py {dev_mode_flag} --port {port} --no-cluster {no_cluster} >> {logfile} 2>&1"
-    else:
-        launch_str = f"{pypath} python app.py {dev_mode_flag} --port {port} >> {logfile} 2>&1"
-
+    update_config({"user_interface": {"port": port}, "dispatcher": {"port": port}})
+    launch_str = f"{pypath} python app.py {dev_mode_flag} --port {port} --no-cluster {no_cluster} >> {logfile} 2>&1"
```
wjcunningham7, line 179: I don't think the `--no-cluster {no_cluster}` is correct

covalent_dispatcher/_cli/service.py
```diff
@@ -320,21 +315,18 @@ def start(
         click.echo(MIGRATION_COMMAND_MSG)
         return ctx.exit(1)
 
+    update_config({"user_interface": {"port": port}, "dispatcher": {"port": port}})
+    if not no_cluster:
+        update_config(
+            {
+                "dask": {
+                    "mem_per_worker": mem_per_worker or "auto",
```
wjcunningham7, line 323: I think `update_config` needs to be called for each of these options individually, if they are specified.  Otherwise specifying one of them will overwrite the config for all of them.

wjcunningham7 (changes requested):

covalent_dispatcher/_cli/service.py
FyzHsn: Right! Let me try that - something about that seemed iffy

PR #965, opened 2022-08-03
Electron list metadata fix

poojithurao:


tests/covalent_tests/workflow/electron_test.py
```diff
@@ -117,3 +117,32 @@ def workflow(x):
...
+
+    workflow.build_graph([2])
+    
+    task_metadata = workflow.transport_graph.get_node_value(0, "metadata")
+    e_list_metadata = workflow.transport_graph.get_node_value(1, "metadata")
+
+    assert list(e_list_metadata["call_before"]) == []
+    assert list(e_list_metadata["call_after"]) == []
+    
+    assert e_list_metadata["executor"] == task_metadata["executor"]
```
cjao, line 148: To make the test a bit more comprehensive, could you set an executor instance for `task`, like
```python3
ct.electron(executor=LocalExecutor())
def task(x, y=0):
```

cjao (changes requested):
This looks good. I'm just requesting a small enhancement to your test.

tests/covalent_tests/workflow/electron_test.py
poojithurao: Done.

cjao (dismissed):
Looks good, thanks for the fix!

scottwn:
pre-commit.ci autofix

scottwn:
pre-commit.ci autofix

scottwn (approved):

cjao:
Follow up observation -- stripping the call `call_before` and `call_after` fields would also remove pip deps. This might be inconvenient if decoding the list elements requires extra pip packages to be installed in the executor.

cjao:
Would it be better to just strip the file transfer calldeps?

PR #966, opened 2022-08-03
Fix config overwriting

### LINKED ISSUE #879: Covalent config is improperly overwritten

FyzHsn:
Fixes #879 


venkatBala (dismissed):

covalent_dispatcher/_cli/service.py
```diff
@@ -320,22 +320,31 @@ def start(
...
-            "dispatcher.address": "localhost",
             "dispatcher.port": port,
-            "dask": {
-                "mem_per_worker": mem_per_worker or "auto",
-                "threads_per_worker": threads_per_worker or 1,
-                "num_workers": workers or dask.system.CPU_COUNT,
-            },
         }
     )
+    if no_cluster_flag not in sys.argv:
```
wjcunningham7, line 330: I think the logic we want is this:
```
if not no_cluster:
    if mem_per_worker != -1:
        set_config({"dask": {"mem_per_worker": mem_per_worker}})
    if threads_per_worker != -1:
        set_config({"dask": {"threads_per_worker": threads_per_worker}})
    if workers != -1:
        set_config({"dask": {"num_workers": workers}})
```

and then for each of those three variables, the click defaults should be set to -1.  The actual defaults that you want in the config should then be modified in the `dask` block inside `covalent/_shared_files/defaults.py`.

FyzHsn: @wjcunningham7 

I checked the `if no_cluster_flag not in sys.argv:` vs `if not no_cluster:` and they both work.

I just added the updated logic according to your comment. Also note that there's some nuance when the defaults haven't been set yet in the config and the user may or may not have specified some default. I'll probably do some refactoring to reduce some of the code bloat here if you approve the logic

covalent_dispatcher/_cli/service.py
```diff
@@ -320,22 +320,70 @@ def start(
...
-            "dask": {
-                "mem_per_worker": mem_per_worker or "auto",
-                "threads_per_worker": threads_per_worker or 1,
-                "num_workers": workers or dask.system.CPU_COUNT,
-            },
         }
     )
 
+    if not no_cluster:
+        if not mem_per_worker:
```
wjcunningham7, line 331: I think this should be `if mem_per_worker`

covalent_dispatcher/_cli/service.py
```diff
@@ -253,7 +253,7 @@ def _graceful_shutdown(pidfile: str) -> None:
     required=False,
     is_flag=False,
     type=str,
-    default="auto",
+    # default="auto",
```
wjcunningham7, line 256: are the defaults "" for strings and 0 for ints?  If so you can remove these commented lines

wjcunningham7 (changes requested):

covalent_dispatcher/_cli/service.py
venkatBala: the default needed for `mem_per_worker` is "auto" and `threads_per_worker =1`, and `num_workers=dask.system.CPU_COUNT`

covalent_dispatcher/_cli/service.py
```diff
@@ -320,22 +320,70 @@ def start(
...
-                "mem_per_worker": mem_per_worker or "auto",
-                "threads_per_worker": threads_per_worker or 1,
-                "num_workers": workers or dask.system.CPU_COUNT,
-            },
         }
     )
 
+    if not no_cluster:
+        if not mem_per_worker:
+            set_config(
```
wjcunningham7, line 332: proper way to call this is `set_config("dask.mem_per_worker", mem_per_worker)`

venkatBala:
@wjcunningham7 I've made the necessary changes as per discussion this morning.

@FyzHsn Kindly review the logic and see if everything is upto spec

FyzHsn:
> @wjcunningham7 I've made the necessary changes as per discussion this morning.
> 
> @FyzHsn Kindly review the logic and see if everything is upto spec

Yessir, the logic is looking good! I'll update the tests to meet the diff coverage requirement.

wjcunningham7 (dismissed):

wjcunningham7 (dismissed):

venkatBala (dismissed):

wjcunningham7 (approved):

PR #974, opened 2022-08-05
Fixing broken tests

kessler-frost:

* Fixing tests that are breaking in develop due to some recent changes.
* Some functional tests are also misplaced into unit tests so those will be moved to their appropriate directories.
* Some unit tests are hitting the live DB so those will be changed to hit an in memory test DB.


kessler-frost:
Also unskipped and fixed some tests hence the slight increase in coverage.

tests/functional_tests/_core/execution_test.py
```diff
@@ -0,0 +1,101 @@
...
+    """Note: This is not a full unit test for the _run_task method. Rather, this is intended to test the diff introduced to write the sublattice electron id in the Database."""
+
+    # class MockResult:
+    #     dispatch_id = "test"
+
+    # def mock_func():
+    #     return MockResult()
+    # class MockSerializedCallable:
+    #     def get_deserialized(self):
+    #         return mock_func
```
FyzHsn, line 72: (non-blocking) Can we delete these commented lines?

FyzHsn (approved):
You get an A+ for mocking and testing! I added one minor non-bocking comment. Up the code coverage and this PR is good ter go!

tests/functional_tests/_core/execution_test.py
FyzHsn: You need to test that one line that you added

kessler-frost: On it.

santoshkumarradha (commented):
Approved ✅

PR #983, opened 2022-08-05
Make `BaseAsyncExecutor` and `BaseExecutor` siblings

### LINKED ISSUE #978: Make `BaseExecutor` and `BaseAsyncExecutor` siblings
This is a followup to PR #896, specifically the [following discussion](https://github.com/AgnostiqHQ/covalent/pull/896#discussion_r927636316)

>What do you think of making BaseAsyncExecutor its own base class instead of a subclass of BaseExecutor? That way we can enforce that run() is a coroutine. Trying to handle the case where run() is not a coroutine here would block the event loop.

>We can make a simple BaseAsyncExecutor base class with just execute and run. All the logic pertaining to a local conda env can be jettisoned as it is incompatible with async.

>Since an async run() is not really "derived" from non-async run() (they aren't invoked in the same way), BaseAsyncExecutor is logically parallel to, not derived from, BaseExecutor.

...

>I would've agreed with that, actually I was going to do it like that at first, but a lot of things in BaseAsyncExecutor are the same as in BaseExecutor which are unrelated to conda, and their respective function signature, attribute types are similar as well; things like: log_stdout, log_stderr, cache_dir, get_dispatch_context (although this will be removed soon), write_streams_to_file(), to_dict(), from_dict(), short_name().

>Since all of these functions/attributes will need to be copy-pasted in the new BaseAsyncExecutor, I figured it was better to derive them from BaseExecutor. Perhaps if we put those things in a common file then both executors can import it from there? Might be out of the scope of this issue.

### LINKED ISSUE #980: Lattice JSON serialization test fails randomly
The lattice JSON serialization test (`test_lattice_json_serialization`) fails randomly. This is because Lattice has an attribute `cova_imports` which is a Python set, and serialization converts the set to a list. However, the list order is not deterministic.

Work around this by setting `cova_imports` to a singleton for the purposes of the test.

### LINKED ISSUE #984: PR tests attempt to validate webapp even if the webapp is not built
The automated tests for a PR were [recently changed](https://github.com/AgnostiqHQ/covalent/blob/614eb2eeeaedf7fa44870614b3f643c1ab233918/.github/workflows/tests.yml#L177) to build the webapp only if files under `covalent_ui/webapp` were changed. However, the "validate distribution" step still [always tries to validate](https://github.com/AgnostiqHQ/covalent/blob/614eb2eeeaedf7fa44870614b3f643c1ab233918/.github/workflows/tests.yml#L205) the `webapp/build` directory.

cjao:

This also introduces a private `_AbstractBaseExecutor` class
containing a few common attributes and methods.

`BaseAsyncExecutor` is logically related to but not derived from
`BaseExecutor` because the core `execute()` methods are invoked
differently.

By not deriving `BaseAsyncExecutor` from `BaseExecutor`, we also
prevent mistakes like the following:

Let `executor` be derived from `BaseAsyncExecutor`.

```python
if isinstance(executor, BaseExecutor):
    # BaseAsyncExecutors are BaseExecutors. But this doesn't work b/c
    # execute() is a coroutine.
    executor.execute()
else:
    await executor.execute()
```

Closes #978

Also addresses the following CI issues: 
Closes #980 (second attempt), 
Closes #984.

### DIFF
--- a/tests/covalent_tests/executor/base_test.py
+++ b/tests/covalent_tests/executor/base_test.py
@@ -28,7 +28,7 @@ from functools import partial
 
 from covalent import DepsCall, TransportableObject
 from covalent.executor import BaseExecutor, wrapper_fn
-from covalent.executor.base import BaseAsyncExecutor
+from covalent.executor.base import BaseAsyncExecutor, _AbstractBaseExecutor
 
 
 class MockExecutor(BaseExecutor):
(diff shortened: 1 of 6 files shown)


cjao:
@scottwn @wjcunningham7 Let me know if you prefer the CI fixes split off into their own PR.

tests/covalent_tests/executor/base_test.py
```diff
@@ -331,3 +331,41 @@ async def fake_run(function, args, kwargs, task_metadata):
...
+
+    me = MockAsyncExecutor()
+
+    # Case 1 - Check that relative log files that are written to are constructed in the results directory that is explicitly passed as an argument.
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        tmp_file = "relative.log"
+        write_awaitable = me.write_streams_to_file(
+            stream_strings=["relative"], filepaths=[tmp_file], dispatch_id="", results_dir=tmp_dir
+        )
+        asyncio.run(write_awaitable)
```
kessler-frost, line 349: Instead of `asyncio.run` you can probably mark this whole test as a `@pytest.mark.asyncio` function, make it `async def` and then `await write_awaitable` here. If that doesn't work, then this is fine too.

tests/covalent_tests/workflow/lattice_serialization_test.py
```diff
@@ -35,7 +35,7 @@ def workflow(x):
         return f(x)
 
     workflow.build_graph(5)
-    workflow.cova_imports = set("dummy_module")
+    workflow.cova_imports = set({"dummy_module"})
```
kessler-frost, line 38: This automatically becomes a set if you just use `{}`, no type conversion needed.

kessler-frost (approved):

scottwn (approved):

PR #1020, opened 2022-08-12
Run workflows directly on the FastAPI event loop

### LINKED ISSUE #1019: Dask executor plugin cannot reliably handle sublattices
The dask executor plugin cannot reliably handle simultaneous dispatches, and fails in particular for many sublattice workflows (e.g. the `vary_distance` workflow in the quantum chemistry tutorial as of v0.177.0rc0). This is due to a confluence of factors:

- Covalent tries to share a single asynchronous Dask Client among all the tasks within a workflow that execute in the same Dask cluster. This client is cached (in the `_address_client_map`) upon initialization by the first task, and retrieved from cache by subsequent tasks.

- Each workflow runs in its own thread in the `workflow_pool`. Because different threads share the same memory space, the same `_address_client_map` is shared by all workflows.

- When one retrieves a result from a Dask `Future`, the Future [checks](https://github.com/dask/distributed/blob/930d3dc023a43990374b5d6a567dfe3ac3dcc428/distributed/client.py#L273) whether the client that submitted the task is operating in `asynchronous` mode. Only when `future.client.asynchronous` is `True` can the future be awaited on.

- `Client.asynchronous` is actually a [property](https://github.com/dask/distributed/blob/930d3dc023a43990374b5d6a567dfe3ac3dcc428/distributed/utils.py#L322) that compares the current event loop with the event loop that created the client. If the client was created from a different thread (in particular, from a different workflow), the `asynchronous` property is False. Thus, the current thread would need to recreate the Client to obtain an awaitable future. But then the `asynchronous` property would be `False` when queried by other threads. This potentially leads to a race condition if the current thread is preempted [between lines 111 and 115](https://github.com/AgnostiqHQ/covalent/blob/d7daa861af08111c644f77432dab648d938ec602/covalent/executor/executor_plugins/dask.py#L115), since the `asynchronous` property could revert to False when [checked in line 115](https://github.com/dask/distributed/blob/930d3dc023a43990374b5d6a567dfe3ac3dcc428/distributed/client.py#L502).

This was less of a problem before because different workflows were dispatched in separate *processes*, not threads, so Dask clients weren't shared between workflows.

**Proposal**: get rid of the `workflow_pool` and run workflows as `asyncio` tasks on the FastAPI event loop. As a side benefit, we can dispense with the [hack](https://github.com/AgnostiqHQ/covalent/blob/d7daa861af08111c644f77432dab648d938ec602/covalent_dispatcher/_service/app.py#L49) for starting the global `workflow_pool` and `tasks_pool`.

### LINKED ISSUE #1021: wait enum conflicts with `ct.wait`
The new `wait` [enum](https://github.com/AgnostiqHQ/covalent/blob/develop/covalent/_results_manager/wait.py) controls how long the client `get_result` waits for the result. Covalent tries to expose that enum as `ct.wait`. Unfortunately, there is [already](https://github.com/AgnostiqHQ/covalent/blob/develop/covalent/__init__.py#L36) an unrelated function `ct.wait`. 

Proposal: rename the enum to `ct.WAIT`.

cjao:

Closes #1019 
Closes #1021


AlejandroEsquivel (approved):
Looks good @cjao 👍

mshkanth:
@mpvgithub @Aravind-psiog @ArunPsiog @Prasy12 - Pls check the changes on this and the bugs that it fixes

PR #1047, opened 2022-08-16
Removing `ct.wait.LONG` from tutorials

### LINKED ISSUE #1035: RTD reference to `ct.wait.LONG` must be removed
## Acceptance Criteria

- RTD reference to `ct.wait.LONG` in MNIST Tutorial must be removed

kessler-frost:
- [x] I have added/updated the tests to cover my changes.

Resolves #1035.


santoshkumarradha (approved):
GTG from docs side.

AlejandroEsquivel (approved):
Thanks @kessler-frost I think the behavior on `get_result` is sensible, although it would be worth having a longer discussion about it's behavior.

As I see it there should be two flavours, one is async with one re-retry, and another completely synchronous with `EXTREME` retries with an additional (timeout in seconds) arg.

Thanks for getting this in !

PR #1053, opened 2022-08-16
Fix release workflow

venkatBala:


wjcunningham7 (approved):

.github/workflows/release.yml
```diff
@@ -308,26 +308,31 @@ jobs:
...
-        run: |
-          docker build -t covalent .
-          docker tag covalent:latest public.ecr.aws/covalent/covalent:latest
-          docker tag covalent:latest public.ecr.aws/covalent/covalent:$TAG
-          if [ "${{ github.events.inputs.test_release }}" == false ] ; then
-            docker push -a public.ecr.aws/covalent/covalent
-          fi
-      - name: Update stable tag
+          github.event_name == 'workflow_call'
+          && github.events.inputs.pre_release == 'true'
```
scottwn, line 320:
```suggestion
          && github.events.inputs.prerelease == 'true'
```

.github/workflows/release.yml
```diff
@@ -308,26 +308,31 @@ jobs:
...
+      - name: Login to Public Registry
+        uses: docker/login-action@v1
+        with:
+          registry: public.ecr.aws
+          username: ${{ secrets.AWS_KEY_ID }}
+          password: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      - name: Build and push pre-release
         if: >
-          inputs.prerelease
-          && ${{ !github.events.inputs.stable_version }}
```
scottwn, line 315: I wouldn't remove this double check

.github/workflows/release.yml
```diff
@@ -308,26 +308,31 @@ jobs:
...
+          context: .
+          file: ./Dockerfile
+          platforms: all
+          push: true
+          tags: |
+            public.ecr.aws/covalent/covalent:latest
+            public.ecr.aws/covalent/covalent:${{ env.TAG }}
+      - name: Tag stable release
         if: >
-          github.event.inputs.stable_version
```
scottwn, line 325: we should keep this; we only want to do the stable release if the `stable_version` exists

scottwn (changes requested):

.github/workflows/release.yml
```diff
@@ -308,26 +308,34 @@ jobs:
...
-        run: |
-          docker build -t covalent .
-          docker tag covalent:latest public.ecr.aws/covalent/covalent:latest
-          docker tag covalent:latest public.ecr.aws/covalent/covalent:$TAG
-          if [ "${{ github.events.inputs.test_release }}" == false ] ; then
-            docker push -a public.ecr.aws/covalent/covalent
-          fi
-      - name: Update stable tag
+          && github.event_name == 'workflow_call'
+          && github.events.inputs.prerelease == 'true'
```
scottwn, line 322: actually maybe we don't need this at all? `inputs.prerelease` is already checked in line 320

.github/workflows/release.yml
```diff
@@ -308,26 +308,34 @@ jobs:
...
+            public.ecr.aws/covalent/covalent:${{ env.TAG }}
+      - name: Tag stable release
         if: >
-          github.event.inputs.stable_version
-          && !inputs.prerelease
-          && steps.docker-build-push.outcome == 'success'
-          && ${{ !github.event.inputs.test_release }}
+          github.event_name == 'workflow_dispatch'
+          && github.event.inputs.stable_version
+          && !github.events.inputs.test_release
```
scottwn, line 337: what's the reason for this syntax change?

scottwn (changes requested):

.github/workflows/release.yml
wjcunningham7: the ${{}} are not needed in `if` clauses

.github/workflows/release.yml
```diff
@@ -308,26 +308,33 @@ jobs:
...
+          file: ./Dockerfile
+          platforms: all
+          push: true
+          tags: |
+            public.ecr.aws/covalent/covalent:latest
+            public.ecr.aws/covalent/covalent:${{ env.TAG }}
+      - name: Tag stable release
         if: >
-          github.event.inputs.stable_version
-          && !inputs.prerelease
```
scottwn, line 326: do we not want to also keep this check?

scottwn (changes requested):

.github/workflows/release.yml
wjcunningham7: @venkatBala please add that line back in

venkatBala: added

venkatBala:
@scottwn this should be good to go thanks!

scottwn (approved):

PR #1144, opened 2022-08-29
enabling frontend test coverage

wjcunningham7:


wjcunningham7:
@scottwn the original purpose of this PR was to get the front-end coverage to start being reported... but I ended up also cleaning up some of the logic in `tests.yml`.  The best way to review this will be to mostly ignore the diff, and just re-read the file fresh.

Notable changes:
- `covalent-dev` images are used so that we can get rid of steps installing NPM and yarn
- The filters are slightly modified and then in a subsequent step they are used to set environment variables. Those variables more accurately convey what we are trying to do in certain scenarios and are used in most of the rest of the workflow.
- Several steps now use `if: steps.xxx.outcome == 'success'` to indicate it depends on a previous step *not being skipped*, i.e., it means it's triggered on the same logical condition.
- I've defined `RECOMMENDED_PLATFORM` as an additional environment variable to allow us to more easily identify which job in the matrix is responsible for uploading coverage reports, pushing to master, and notifying Slack.

.github/workflows/tests.yml
```diff
@@ -84,107 +84,73 @@ jobs:
...
+            NEED_PYTHON=1
+          fi
+          NEED_FRONTEND=0
+          if [ ${{ steps.modified-files.outputs.ui_frontend }} = 'true' ] ; then
+            NEED_FRONTEND=1
+          fi
+          BUILD_AND_RUN_ALL=0
+          if [ ${{ github.event_name }} != 'pull_request' ] \
+            || [ ${{ steps.modified-files.outputs.build }} = 'true' ] \
+            || [ ${{ steps.modified-files.outputs.actions }} = 'true' ] ; then
```
scottwn, line 129:
```suggestion
            || [ ${{ steps.modified-files.outputs.build }} = 'true' ] \
            || [ ${{ steps.modified-files.outputs.functional_tests }} = 'true' ] \            
            || [ ${{ steps.modified-files.outputs.actions }} = 'true' ] ; then
```

scottwn (changes requested):

.github/workflows/tests.yml
```diff
@@ -84,107 +84,73 @@ jobs:
...
             - 'MANIFEST.in'
-            - 'requirements.txt'
+            - '**/requirements.txt'
             - 'setup.py'
             actions:
-            - '.github/workflows/*.yml'
-            tests:
-            - 'tests/**'
+            - '.github/workflows/*'
+            - '.github/actions/**'
```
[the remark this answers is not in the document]
wjcunningham7: done

.github/workflows/tests.yml
wjcunningham7: done

scottwn (approved):

PR #1151, opened 2022-08-30
prototyping prod docker image

wjcunningham7:


wjcunningham7:
The goal of this change is to provide a Dockerfile which builds quickly and can be used in a self-hosted deployment setting.

Summary of changes:
- The existing Dockerfile is primarily geared towards development, i.e., always building Covalent from source. This suggested the need for a different (production) Dockerfile which can be used in production, based on a stable release.
- `rsync` and `wget` installations are moved to the build stage, with relevant binaries/libraries copied to the prod stage
- Container is run as `ubuntu` instead of as `root` to improve security
- Many labels have been added
- Healthcheck added and working. Takes about 30 sec to go from starting to healthy
- Results directory modified in the config, otherwise the server crashes.  This suggests outstanding work will primarily involve dealing with client-server results communication.  Note that currently dispatches still fail because the client's results directory is injected into the request.
- Server is started using the CLI instead of using `python`.

wjcunningham7:
At a later point when this is all working, I'd like the CI's release stage to use this rather than re-building Covalent, similar to how the `conda` job works.

scottwn (approved):

PR #1175, opened 2022-09-06
fixing recommended platform

### LINKED ISSUE #1164: Problems with conditional logic in tests.yml
### Environment

- **Covalent version**: 0.190.0
- **Python version**: 3.8
- **Operating system**: Ubuntu 18.04

### What is happening?

environment variables in tests.yml are not being set correctly

I believe the correct "recommended platform" is Ubuntu 18.04 / Python 3.8 / Dask

the matrix.container for this platform is `ghcr.io/agnostiqhq/covalent-dev/ubuntu18-py38:latest`

however the workflow is looking for `python:3.8-bullseye`, so `RECOMMENDED_PLATFORM` is always empty

also, there's a double `==` where there should be a single `=`
https://github.com/AgnostiqHQ/covalent/blob/6ecbd17361d93762a1acb63b82849dfdff035a6e/.github/workflows/tests.yml#L126

Open any PR with passing tests, like #1157 

### What should happen?

`RECOMMENDED_PLATFORM` should be `true` for  Ubuntu 18.04 / Python 3.8 / Dask

### Any suggestions?

match on `ghcr.io/agnostiqhq/covalent-dev/ubuntu18-py38:latest` instead of `python:3.8-bullseye`

wjcunningham7:

Fixes #1164


wjcunningham7:
@cjao if the codecov report is generated here then we'll know it was resolved.

cjao:
It does appear to be resolved :D

cjao (commented):
Codecov is generated again but there is still the double `==` in line 126.

scottwn (approved):

PR #1180, opened 2022-09-06
Refactor sublattice handling

### LINKED ISSUE #1170: Return sublattice transport graph directly to the dispatcher
The logic for dispatching sublattices needs to be refactored to allow associated metadata to be written to the DB at the same time as the sublattice itself.

The dispatcher's [current approach to handling sublattices](https://github.com/AgnostiqHQ/covalent/blob/15b6e280fac69ecb6d3fee020da2ff9c668b883e/covalent_dispatcher/_core/execution.py#L198) is to send a `_dispatch_sublattice` job to the workflow_executor. This job performs an out-of-process ct.dispatch() on the sublattice workflow function, which in turn builds the sublattice transport graph and submits the sublattice to the dispatcher's `/api/submit` endpoint. This causes the dispatcher to write the received sublattice and its transport graph to the DB, call run_workflow asynchronously, and return the sublattice dispatch id to the execution of the parent lattice. Only at this point is the `electron_id` property is [attached](https://github.com/AgnostiqHQ/covalent/blob/efb2ab2ecd78b8cb6cac9654efebd4bc07fcc6d7/covalent_dispatcher/_core/execution.py#L299) to the sublattice db record. 

**Problems with this approach**
* Although `electron_id` is already known when the sublattice is dispatched, it is only written to the DB some time afterwards. **The sublattice and its parent lattice are unrelated in the DB during this window**; the sublattice record would be irrecoverably orphaned if the execution is interrupted.

* To support an upcoming UI feature (see #1159) we will also need to add a `root_lattice_dispatch_id` property to each lattice and sublattice. Writing `root_lattice_dispatch_id` after the sublattice is dispatched will not work; the information might be needed before that write occurs.

Instead of calling `/api/submit`, the workflow_executor should simply return the built-out sublattice. The dispatcher can then recover the sublattice and its transport graph via the `TransportableObject.json` attribute and write the entire sublattice record atomically.

Acceptance criteria:
- [ ] Add `root_dispatch_id` to `Result`
- [ ] Modify `Result.persist()` to write `electron_id` in the same transaction as the main Lattice record; the `root_dispatch_id` attribute will be handled by #1171.
- [ ] Modify handling of sublattices in `execution.py` as outlined above.

cjao:

Closes #1170.


covalent_dispatcher/_core/execution.py
```diff
@@ -57,15 +56,36 @@
...
+    if parent_result_object:
+        result_object._root_dispatch_id = parent_result_object._root_dispatch_id
+
+    result_object._initialize_nodes()
+    app_log.debug("2: Constructed result object and initialized nodes.")
+
+    result_object.persist(electron_id=parent_electron_id)
+    app_log.debug("Result object persisted.")
+
+    return result_object
```
scottwn, line 88: #1174 is merged, do you want to move this?

scottwn (changes requested):

covalent_dispatcher/_core/execution.py
cjao: Good point, I have moved this.

scottwn (approved):

PR #1190, opened 2022-09-08
Fix file transfer for collected/generated electrons

### LINKED ISSUE #1179: Make File Transfer compatible with electrons with iterable outputs
### Environment

- **Covalent version**: 0.177.0
- **Python version**: 3.8.13
- **Operating system**: mac OS

### What is happening?

When transferring files and returning the results as a list / tuple / dict, we get an error saying: 

```
[2022-09-06 19:25:18,248] [ERROR] execution.py: Line 372 in _run_task: Run task exception
  File "/Users/faiyaz/opt/anaconda3/envs/dnn-tutorial/lib/python3.8/site-packages/covalent_dispatcher/_core/execution.py", line 345, in _run_task
    output, stdout, stderr = await execute_callable()
  File "/Users/faiyaz/opt/anaconda3/envs/dnn-tutorial/lib/python3.8/site-packages/covalent/executor/base.py", line 572, in execute
    result = await self.run(function, args, kwargs, task_metadata)
  File "/Users/faiyaz/opt/anaconda3/envs/dnn-tutorial/lib/python3.8/site-packages/covalent/executor/executor_plugins/dask.py", line 112, in run
    result, worker_stdout, worker_stderr = await dask_client.gather(future)
  File "/Users/faiyaz/opt/anaconda3/envs/dnn-tutorial/lib/python3.8/site-packages/distributed/client.py", line 2037, in _gather
    raise exception.with_traceback(traceback)
  File "/Users/faiyaz/opt/anaconda3/envs/dnn-tutorial/lib/python3.8/site-packages/covalent/executor/executor_plugins/dask.py", line 60, in dask_wrapper
    output = fn(*args, **kwargs)
  File "/Users/faiyaz/opt/anaconda3/envs/dnn-tutorial/lib/python3.8/site-packages/covalent/executor/base.py", line 95, in wrapper_fn
    output = fn(*new_args, **new_kwargs)
TypeError: get_item() got an unexpected keyword argument 'files'
```

The UI shows the following:
Screen Shot 2022-09-06 at 3 25 33 PM

Add a train.csv and test.csv in the login node in beehive under your account. Accordingly replace the filepaths in the example workflow shown below.

```
remote_train_file = ct.fs.File('/federation/faiyaz/source_data/train.csv', is_remote=True)
remote_test_file = ct.fs.File('/federation/faiyaz/source_data/test.csv', is_remote=True)
local_train_file = ct.fs.File('/Users/faiyaz/Code/datasets/train.csv')
local_test_file = ct.fs.File('/Users/faiyaz/Code/datasets/test.csv')

strategy = ct.fs.strategies.Rsync(
    user='faiyaz',
    host='beehive.agnostiq.ai', 
    private_key_path='/Users/faiyaz/.ssh/id_ed25519'
)

@ct.electron(
    files=[
        ct.fs.FileTransfer(remote_train_file, local_train_file, strategy=strategy, order=ct.fs.Order.BEFORE),
        ct.fs.FileTransfer(remote_test_file, local_test_file, strategy=strategy, order=ct.fs.Order.BEFORE),
    ]
)
def fetch_and_validate_data(files=[]) -> List[str]:
    _, local_train = files[0]
    _, local_test = files[1]
    train = pd.read_csv(local_train, parse_dates=['date'])
    test = pd.read_csv(local_test, parse_dates=['date'])
    return train, test

@ct.lattice
def workflow():
    train, test = fetch_and_validate_data()
    
print(res)
```

### What should happen?

The iterables should be unpacked without any errors. Also, note that for a non-iterable return, there is no issue.

### Any suggestions?

It's possible that during L218 of electron.py, particularly in `def __iter__()` when consructing the electron we copy the metadata `self.metadata.copy()` which would also include the callbefore & callafter deps with return value keys defined, which would in turn try to inject the `files` kwarg. We may want to disable this for internally generated electrons or figure out a way to make the injected keywords optional.

FyzHsn:
Fixes #1179 


covalent/_workflow/electron.py
```diff
@@ -238,7 +238,13 @@ def get_item(e, key):
 
                 get_item.__name__ = node_name
 
-                get_item_electron = Electron(function=get_item, metadata=self.metadata.copy())
+                iterable_metadata = self.metadata.copy()
+                iterable_metadata[
```
cjao, line 242: Hi @FyzHsn, unfortunately this approach is a bit too blunt. For example, what would happen if a pip dep is needed to interpret the inputs to `get_item`? See #1191.

cjao (changes requested):

cjao (approved):
As discussed yesterday, this hack does fix the immediate problem, but we should filter out all file transfers in these auto-generated electrons in the near future (see #1191).

kessler-frost:
Hmm, I'm a bit concerned and confused. If the problem is with a `generator` electron, then why is it even executed? And if it isn't executed then how does its metadata matter? Shouldn't it be only handled in post processing?

cjao:
Generator electrons work a bit differently from before. The unpacking happens in an executor because the dispatcher would generically not have the dependencies to unpack an encoded collection. This is analogous to how collection-type electrons work now.

kessler-frost:
I see, so does that mean these electrons will also get executed in a separate executor? (If yes, then we definitely need to change that).

kessler-frost:
Whew, ok, so @FyzHsn I would say the problem is a lot more fundamental than I thought it was. I assume that is what you were trying to convey earlier.

kessler-frost:
>The unpacking uses the same executor as the main electron. That is guaranteed to work because the main electron's executor will have the dependencies to generate the collection that needs to be unpacked.

Yeah, but that means it'll occupy a huge amount of resource to do an extremely trivial thing which theoretically can be handed off to the electron which actually needs it.

kessler-frost:
Another solution might be to add tags to the `TransportableObject` which will help in filtering these things out on the destination electron's remote machine itself.

cjao:
> > The unpacking uses the same executor as the main electron. That is guaranteed to work because the main electron's executor will have the dependencies to generate the collection that needs to be unpacked.
> 
> Yeah, but that means it'll occupy a huge amount of resource to do an extremely trivial thing which theoretically can be handed off to the electron which actually needs it.

Postprocessing and the collection-type electrons have the same problem. Ideally we could just reserve the main electron's executor until the auxiliary tasks have finished. The planned [executor-sharing](https://github.com/AgnostiqHQ/covalent/discussions/943#) feature could help.

cjao (dismissed):
Looks good! Followup discussion in #1196 .

kessler-frost (approved):

cjao (approved):

Issue #908, opened 2022-07-25
Restore ability to query result object properties from memory


cjao:
During the initial migration to the new DB schema in #823, several methods of `Result` to query the transport graph were modified to always look up the data in the db. These are:

* `get_node_result()`
* `get_all_node_results()`
* `get_all_node_outputs()`
* `_get_node_name()`
* `_get_node_status()`
* `_get_node_output()`
* `_get_node_error()`

We will need to query the db for this information once we add the ability to resume workflows as the server would then need to rehydrate a partially executed computational graph. Presently, however, hitting the db is not yet necessary; the dispatcher  holds the entire transport graph in memory during execution. Further, these methods are also exposed to the client which also has the entire Result object in memory but doesn't have access to the server's db.

scottwn:
@wjcunningham7 I believe this is out of scope until we support resuming workflows, is that correct?

scottwn:
`get_node_result()` relies on `get_node_name()`, `get_node_output()`, etc., which were updated to use the in memory `Result` in #1216

`get_all_node_results()` in turn relies on `get_node_result()`

`_get_node_output()` was changed to get the output from the in memory `Result` in https://github.com/AgnostiqHQ/covalent/commit/9e321a9a97ea82a816db95b2c62b09cd00366686. `get_all_node_outputs()` relies on `_get_node_output()`.

`_get_node_status()` was changed to get status from the in memory `Result` in https://github.com/AgnostiqHQ/covalent/commit/32929263ab5defc3232f72b8da2c2fe8facf33f9

Functionality to retrieve the error from the in memory `Result` was added in https://github.com/AgnostiqHQ/covalent/commit/32929263ab5defc3232f72b8da2c2fe8facf33f9

Functionality to retrieve the node name from the in memory `Result` was added in https://github.com/AgnostiqHQ/covalent/commit/32929263ab5defc3232f72b8da2c2fe8facf33f9

PR #1192, opened 2022-09-08
Update tests for v12

cjao:


cjao:
This is going to require some more additional work.  It turns out that many of the DBs in `execution_test.py` aren't mocked correctly, so the tests are actually hitting a DB on disk. I only found out after moving `test_results_post_process` b/c the real DB isn't initialized for the sdk tests. 

Also, I  think the integration tests questions can be decided later. For the time being, if the main goal is to remove the dependency on the DB, I'm thinking of just refactoring all the result object methods like `Result._update_node` to isolate the sqlalchemy commands into mockable functions. I'm reluctant to mock `_run_planned_workflow` entirely because the bugs that prompted `test_run_workflow_with_failing_nonleaf` and `test_run_workflow_with_failing_leaf` would not have been caught by the unit tests alone. The different types of node failures are handled in different code paths of `_run_planned_workflow`.

Marking as draft for now.

cjao:
Hi @scottwn, I touched some other execution tests in that PR just to fix the db mocking so that they don't try to hit a persistent db, but those weren't originally in the scope of the PR. Mocking out the DB entirely for those tests would be considerably more involved and amount to fixing [#908](https://github.com/AgnostiqHQ/covalent/issues/908).

cjao:
Hi @scottwn, I touched some other execution tests in that PR just to fix the db mocking so that they don't try to hit a persistent db, but those weren't originally in the scope of the PR. Mocking out the DB entirely for those tests would be considerably more involved and it would probably be easier to just fix [#908](https://github.com/AgnostiqHQ/covalent/issues/908).

scottwn:
This is looking pretty good, can you also create a `MockDataStore` class in app_test.py so that `test_db()` isn't returning a `DataStore` that depends on the schemas defined on the SDK side? The `MockDataStore` can have a very trivial schema, just for testing that "results" are returned as expected.

cjao:
Sorry I missed that last part! Will add it shortly.

cjao:
@scottwn Have a look at the all-new MockDataStore : ).

scottwn (approved):

Issue #1197, opened 2022-09-09
changelog.yml is failing, blocking releases


wjcunningham7:
### Environment

- **Covalent version**: > 0.197.0
- **Python version**: N/A
- **Operating system**: N/A

### What is happening?

See these two runs:
https://github.com/AgnostiqHQ/covalent/actions/runs/3016482600
https://github.com/AgnostiqHQ/covalent/actions/runs/3017070175

### How can we reproduce the issue?

Merge to develop with a change that should trigger a version increment.

### What should happen?

Changelog should be reformatted and version incremeted.

scottwn:
@wjcunningham7 this is probably unrelated, but why are both Ubuntu 18 and Ubuntu 20 `ubuntu-latest`?
https://github.com/AgnostiqHQ/covalent/blob/c73822836e03759a4707d817890f35a41dfa2d7f/.github/workflows/test_matrix.json#L60-L82

scottwn:
per https://github.com/actions/runner/issues/1733 I believe the problem is caused by passing the blob of JSON commits to the action as a var

PR #1198, opened 2022-09-09
#1197 use octokit inside action to get commits

scottwn:


scottwn:
The code in `dist` is automatically generated and can be ignored.

In `action.yml` I've removed commits as an input since Node errors if there are too many commits being passed as an argument. I've added a token and a "basehead" formatted comparison so the action can retrieve commits on its own.

.github/actions/changelog/index.js
```diff
@@ -32,107 +33,131 @@ const get_author = (commit, authors) => {
...
+  octokit.rest.repos
+    .compareCommitsWithBasehead({
+      owner: owner,
+      repo: repo,
+      basehead: basehead,
+    })
+    .then((compareCommits) => {
+      const {
+        data: { commits: commits },
+      } = compareCommits;
```
scottwn, line 54: This is the only meaningful change to the JS. The rest of the diff is indentation and minor changes in variable names.

scottwn:
I've also made corresponding changes in `changelog.yml` to pass the correct vars to the action.

wjcunningham7 (dismissed):

wjcunningham7 (approved):

PR #1245, opened 2022-09-15
Fix benchmark/release workflow

venkatBala:


.github/workflows/release.yml
```diff
@@ -391,6 +391,7 @@ jobs:
           && !inputs.prerelease
           && !github.events.inputs.test_release
           && contains(env.PAUL_BLART, github.actor)
+          && github.event_name != 'schedule'
```
scottwn, line 394: this is unnecessary, we're already checking that the `event_name` is `workflow_dispatch`

scottwn (changes requested):

wjcunningham7 (dismissed):

scottwn (approved):

PR #1251, opened 2022-09-16
Lock config file during config updates

### LINKED ISSUE #1235: Corrupted config file
### Environment

- **Covalent version**: 0.198.0
- **Python version**: 3.8.13
- **Operating system**: MacOS 12.3.1

### What is happening?

It seems the corrupted config file issue has crept up again. This is most likely due to performing multiple writes from different threads.

After installing covalent, when I do

```bash
covalent start --ignore-migrations -d
covalent purge -y
```

Then the following error is thrown:
```bash
toml.decoder.TomlDecodeError: Found invalid character in key name: ']'. Try quoting the key name. (line 33 column 16 char 927)
```

This happens even when say a test is run after starting covalent server. So it is rather sporadic, and might be difficult to reproduce.

### What should happen?

No error should be thrown.

### Any suggestions?

I'd say the solution for this would be to not perform any writes whenever the `ConfigManager` is initialized. Instead do it when necessary. Reading however can still be performed.

cjao:

Closes #1235.


cjao:
Codecov reports 94.81% overall patch coverage whereas the only changed file config.py has 100% patch coverage.

Emmanuel289 (approved):

scottwn (approved):

scottwn (changes requested):
Patch coverage must be 100%

scottwn (approved):

scottwn:
sorry I'm not sure what's going on with codecov

PR #1252, opened 2022-09-16
Fix codecov reporting

wjcunningham7:


wjcunningham7:
Logic change summary:
- On a push to develop that's not invoked by `changelog.yml`, run all tests and upload reports to codecov.  This keeps `develop` up to date with codecov.
- On a pull request, if no reports were uploaded to codecov, invoke our "passthrough" workflow with the same name.  This should satisfy the branch protections as described [here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks).

wjcunningham7:
Many files were changed by the pre-commit hooks in this PR.  The files with changes relevant to codecov are `codecov.yml` and `tests.yml`.

wjcunningham7:
Example of a successful workflow run: https://github.com/AgnostiqHQ/covalent/actions/runs/3072014384

scottwn:
something's not right with pre-commit here... it shouldn't be changing `"` to `'` in the yaml files

wjcunningham7:
> something's not right with pre-commit here... it shouldn't be changing `"` to `'` in the yaml files

I modified `.pre-commit-config.yaml` to do that.  Otherwise it was doing the opposite by default, and messing up my syntax.

scottwn:
I'm not sure that's the right approach here. If it's just an issue of syntax in a single line, we shouldn't change the prettier setting for the whole repo, we should just tell prettier to ignore that line. I'm also curious where prettier is breaking the syntax, because it really shouldn't do that.

wjcunningham7:
> I'm not sure that's the right approach here. If it's just an issue of syntax in a single line, we shouldn't change the prettier setting for the whole repo, we should just tell prettier to ignore that line. I'm also curious where prettier is breaking the syntax, because it really shouldn't do that.

In `tests.yml` in the 'paths-ignore` key, the values have their single quotes switched to double quotes by prettier.

wjcunningham7:
@scottwn cleaned up now

scottwn (dismissed):

PR #1258, opened 2022-09-20
adding graceful shutdown prototype

wjcunningham7:

### DIFF
--- a/covalent_ui/api/main.py
+++ b/covalent_ui/api/main.py
@@ -21,7 +21,7 @@
 """Fastapi init"""
 
 import socketio
-from fastapi import FastAPI, HTTPException, Request
+from fastapi import FastAPI, HTTPException, Request, Query
 from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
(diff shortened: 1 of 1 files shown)


wjcunningham7:
@AlejandroEsquivel does this make sense to include this functionality?  The original idea was to make sure the in-memory `Results` objects are all flushed to the database when the process is killed.  I think cancellation can achieve that.

AlejandroEsquivel:
@wjcunningham7 I think this is an excellent addition as it transitions covalent from the indeterminate state it was during shutdown (before) to one that is well defined (workflow states being cancelled) hence may play well if & when resuming workflows is also addressed.

When we are in the distributed setting (cloud solution) we may want to extend the shutdown signal handler to also transition health check endpoints to return `503`s in advance so that no new connections are sent to the containers serving API requests during rolling updates and that we have time for operations such as this to complete (and processing last active connections). There are also other minor considerations but they won't be a concern for now (some are addressed in Details section of [this module](https://www.npmjs.com/package/express-graceful-exit)).

As for now in a monolithic setting I think uvicorn / starlette do all the heavy lifting around actually closing the http & websocket connections gracefully and believe it will wait until the shutdown handler is complete until actually terminating the process but we may want to verify that by testing that workflow cancellation actually completes by sending `SIGINT` signal to the process running covalent.

PR #1267, opened 2022-09-22
Remove DB dependency from sync

scottwn:


cjao (approved):

FyzHsn (approved):

covalent/_results_manager/results_manager.py
```diff
@@ -297,10 +294,9 @@ def sync(
         for d in dispatch_id:
             _get_result_from_dispatcher(d, wait=True, status_only=True)
     else:
-        with Session(db.engine) as session:
-            rows = session.query(Lattice).all()
-        for row in rows:
-            _get_result_from_dispatcher(row.dispatch_id, wait=True, status_only=True)
+        raise Exception(
```
FyzHsn, line 297: You could add a test to ensure that the exception is raised.

cjao:
Codecov bug should be fixed by #1252

FyzHsn (changes requested):

covalent/_results_manager/results_manager.py
cjao: @FyzHsn I think that is done [here](https://github.com/AgnostiqHQ/covalent/pull/1267/files?diff=unified&w=0#diff-276d9d67cbc2e64e8515b3454d5cfaa07f055965bcc77a194cf24040c8d51eeaR141).

FyzHsn: Codecov seems to think this line is not really invoked. Do you know why that could be?

scottwn: must be something wrong with the codecov reporting? it must be comparing it against the tests in `develop` rather than the tests in this branch. @wjcunningham7 any thoughts?

FyzHsn: Can you check that the Exception msg in the test and the raised Exception match up? One possibility is that a different Exception is getting raised somewhere else and not necessarily the one in line 297.

cjao: @scottwn Codecov is not wrong. I just tried pytest locally. I think the problem is that `result_from(lattice_record)` raises an `UnboundLocalError` exception. Changing the test to `sync(1)` covers that line.

PR #1271, opened 2022-09-23
Redesigning Statuses' implementation

### LINKED ISSUE #1243: Update `Status` dataclass and replace `RESULT_STATUS` class
The `Status` dataclass is currently imported and used in `Result` where all the statuses are defined. Now, since we're going to have new statuses, we need to come up with better definitions than in `covalent/_shared_files/util_classes.py`. 

Acceptance Criteria:

- [x] Have newly defined statuses (which will replace `RESULT_STATUS` class)
- [x] Remove references to `RESULT_STATUS` and instead use workflow/electron level statuses everywhere
- [x] All major electron statuses are to be defined in this electron statuses class
- [x] Will also need to figure out a way to swap out the status definitions depending on the executor

kessler-frost:

Resolves #1243.


covalent/_shared_files/util_classes.py
```diff
@@ -19,34 +19,52 @@
...
+        value = pickle.dumps(value)
+        try:
+            self.put(value, timeout=0.5)
+        except queue.Full:
+            self.get_nowait()
+            self.put_nowait(value)
 
+    def retrieve(self) -> Any:
+        try:
+            value = self.get(timeout=0.5)
```
cjao, line 49: Hi @kessler-frost, the `status_listener` would insert 0.5-second pauses in the event loop b/c `MPQ.get(timeout)` blocks the entire process.

cjao (changes requested):
Hi @kessler-frost, I think using an MPQ for communication between executor and dispatcher would cause the status listeners to block the event loop for 0.5 seconds at a time. For communicating with Async executors an `asyncio.Queue` would avoid this blocking. The non-Async case would be trickier however

covalent/_shared_files/util_classes.py
kessler-frost: Yeah, that's true, my bad. I am now using a `queue.Queue` instead with `get_nowait()` and `put_nowait()` which should resolve this. Since `queue.Queue`'s are inherently threadsafe, this should better for our purpose (instead of using MPQ which also ensure process-safety).

kessler-frost: After our discussion, this has been now switched to an asyncio queue with added threadsafe saving and retrieving methods.

covalent/_shared_files/util_classes.py
```diff
@@ -19,34 +19,55 @@
...
+        self.event_loop.call_soon_threadsafe(self.put_nowait, value)
+
+    def get_nowait_safe(self) -> Any:
+        return self.event_loop.call_soon_threadsafe(self.get_nowait, ())
 
+    def save(self, value: Any) -> None:
+        try:
+            self.put_nowait_safe(value)
+        except asyncio.QueueFull:
+            self.get_nowait_safe()
```
cjao, line 46: I'm a bit worried about a race condition here. Let the dispatcher be running on thread 0 and the (non-async) `executor.execute()` be running in thread 1. 
Consider the following scenario:

0) SafeVariable is already set. The status listener in Thread 0 calls `SafeVariable.get()`.
1) In `SafeVariable.get()`, [line 155](https://github.com/python/cpython/blob/455b494386678f7f20c0cd43407e689f14afac15/Lib/asyncio/queues.py#L155) `if self.empty()`is evaluated; the queue is not empty.
2) Thread 0 is preempted by the OS; Thread 1 starts running.
3) Thread 1 calls `SafeVariable.save()`. Since the queue is nonempty, the `QueueFull` exception is raised, and `self.get_nowait_safe()` pops the value from the queue. 
4) Control switches back to Thread 0. Since `self.empty()` already evaluated to False in step 1, the interpreter jumps to [line 174](https://github.com/python/cpython/blob/455b494386678f7f20c0cd43407e689f14afac15/Lib/asyncio/queues.py#L174) `return self.get_noawait()`. This raises an unhandled `asyncio.QueueEmpty` exception.

kessler-frost: After our discussion, it seems the better way would be to make this back to `queue.Queue` to avoid race conditions, and solve the problem of frequent calls to `get` inside the listener loop.

kessler-frost:
Putting this on hold as we go through its design process.

wjcunningham7:
@kessler-frost can you close this if not in near-term scope

kessler-frost:
Closing as not in near-term scope.

PR #1304, opened 2022-09-30
UI settings feature

Aravind-psiog:


Prasy12 (approved):

AlejandroEsquivel:
@Prasy12 can you also let me know which node version is being used for development of the UI? this shouldn't block this PR but I will ask that we eventually add an `.nvmrc` file in order to pin the node version

covalent_ui/app.py
```diff
@@ -105,7 +107,7 @@ def get_home(request: Request, rest_of_path: str):
         "app:fastapi_app",
         host=host,
         port=port,
-        debug=DEBUG,
+        debug=log_to_file,
```
AlejandroEsquivel, line 110: @Prasy12 is there a reason why we're enabling debug on uvicorn if we enable sdk logging?

covalent_ui/webapp/src/components/settings/SettingsCard.js
```diff
@@ -0,0 +1,1062 @@
...
+        formattedName = name.charAt(0).toUpperCase() + name.slice(1)
+      }
+    }
+    return formattedName
+  }
+
+  const getLabelName = (name) => {
+    let formattedName = name
+    const uSpliit = name.includes('_')
+    if (uSpliit) {
```
AlejandroEsquivel, line 180: @Prasy12 there's an issue with this logic treating for example: `threads_per_worker` would end up having a formattedName of `Threads per_worker`, also there is code duplication of this doing the same thing in `getSubmenuName`.

I would suggest the following code is used to handle that case:

```
const formatUnderscoreConcatenatedString = (str) => {
    return str.split('_').map(word => word.charAt(0).toUpperCase() + word.slice(1)).join(" ")
  }
```

And this be re-used in `getLabelName` and `getSubmenuName`.

Also I would ask that variable naming be improved as `a` and `b` aren't very verbose.

covalent_ui/webapp/src/components/settings/SettingsCard.js
```diff
@@ -0,0 +1,1062 @@
...
+      var a = name
+        .replace(name.at(0), name.at(0).toLocaleUpperCase())
+        .substring(0, name.indexOf('_'))
+      var b = name
+        .replace(name.at(0), name.at(0).toLocaleUpperCase())
+        .substring(name.indexOf('_') + 1, name.length)
+      formattedName = a + ' ' + b.replace(b.at(0), b.at(0).toLocaleUpperCase())
+    } else {
+      if (name === 'slurm') {
```
AlejandroEsquivel, line 166: For any string literals used let's create a seperate file that define these as constants grouped by their usage.

e.g

```
const CONFIG_SECTIONS = {
  SLURM: 'slurm'
}
```

such that this can be re-used as `CONFIG_SECTIONS.SLURM` and similarly for the other config sections

AlejandroEsquivel (changes requested):
Looks good @Prasy12 just some minor changes needed

AlejandroEsquivel:
@Prasy12 let's also add an `.nvmrc` file that pins the node version to the one being used for development, thanks

Prasy12:
@AlejandroEsquivel We have rectified all the points that you have mentioned above and updated the changes as well. Please review and let us know if this works good. We can merge this once you review. Thanks.

covalent_ui/app.py
Prasy12: Fixed

AlejandroEsquivel (approved):

PR #1314, opened 2022-10-04
Added timestamps in fastapi app logs

Emmanuel289:


venkatBala (approved):

Prasy12:
@Emmanuel289 @venkatBala We had already enabled timestamp for the fastapi logs as part of our last logs PR, is there any reason for these new changes? We have some logic written on our UI as well based on this and might break it. 


cc : @santoshkumarradha @mshkanth @mpvgithub @Aravind-psiog

Prasy12:
@Emmanuel289 @venkatBala
Log format that is there for AQ backend logs
image

Log format introduced on fastapi by Psiog to match this format
image


We are showing logs as part of the GUI now and we might need to have a consistent format for us to break these into **datetime**, **status** and **messages**. We are using regex to split the logs into this format. 
Sample UI screenshot
image


With the current change made in this PR, we might have differences in format between the current AQ backend logs and the Fastapi logs which will make it hard for us to break them into the desired format for GUI usage as there are few extra parameters in the new format introduced.
Difference in log format :
image

cc : @Aravind-psiog @mpvgithub @mshkanth @santoshkumarradha

Emmanuel289:
> @Emmanuel289 @venkatBala We had already enabled timestamp for the fastapi logs as part of our last logs PR, is there any reason for these new changes? We have some logic written on our UI as well based on this and might break it.
> 
> cc : @santoshkumarradha @mshkanth @mpvgithub @Aravind-psiog

@Prasy12 Thanks for following this PR. Are you referring to the PR for the [UI settings feature]? (https://github.com/AgnostiqHQ/covalent/pull/1304). I see that [log_handler.py](https://github.com/AgnostiqHQ/covalent/pull/1304/files#diff-0931c96f6a967f134ce97f46e44029e648bc280cd432eb308bb8adbf40559544) that contains some configuration was included in the PR.

Aravind-psiog:
Thanks for the clarification @AlejandroEsquivel l . By default, logging is disabled and the log level is set to warning in the configuration. It can be enabled in the settings section on the GUI under SDK and by setting the log level to info, you can see the logs on UI, log file, and also by running the command ```covalent logs```. Note that covalent should be restarted after making changes in the settings UI since log-level changes apply at the start of Fast API. @Emmanuel289

Emmanuel289:
> Thanks for the clarification @AlejandroEsquivel . By default, logging is disabled and the log level is set to warning in the configuration. It can be enabled in the settings section on the GUI under SDK and by setting the log level to info, you can see the logs on UI, log file, and also by running the command `covalent logs`. Note that covalent should be restarted after making changes in the settings UI since log-level changes apply at the start of Fast API. @Emmanuel289

Thanks for your feedback @Aravind-psiog

AlejandroEsquivel:
It seems that FastAPI logging works with timestamps as long as `sdk.enable_logging` config is enabled. 
Hence this update is not required, thanks all!

PR #1329, opened 2022-10-11
UI terminal feature

Prasy12:

This feature contains ability to use the terminal on the GUI


Prasy12:
@AlejandroEsquivel Can you please review this and help us merge it?


cc: @santoshkumarradha @mshkanth

covalent_ui/webapp/src/components/terminal/Terminal.js
```diff
@@ -0,0 +1,67 @@
...
+
+import React, { useRef, useEffect, memo } from 'react'
+import { XTerm } from 'xterm-for-react'
+import { FitAddon } from 'xterm-addon-fit';
+import io from 'socket.io-client'
+import './index.css';
+const Terminal = () => {
+    const xtermRef = useRef()
+    const fitAddon = new FitAddon();
+    const socket = io(process.env.REACT_APP_SOCKET_URL, {
```
AlejandroEsquivel, line 11: This would get called on re-renders, however I believe the docs say that the connection manager is re-used but I would double check this won't have any side effects https://socket.io/docs/v4/client-api/#iourl otherwise placing in useEffect -> set manager with useState might make sense

covalent_ui/webapp/src/components/terminal/Terminal.js
```diff
@@ -0,0 +1,67 @@
...
+                    xtermRef?.current?.terminal.focus();
+                    socket.emit('start_terminal', dims);
+                    socket.emit("resize", dims);
+                }
+            })
+        }
+    })
+    useEffect(() => {
+        if (socket) {
+            return () => {
```
AlejandroEsquivel, line 31: In the case that the browser is closed and the `stop_terminal` websocket message is not sent, will this create problems with interacting with the terminal when re-opening the UI ? If so, it may be a good idea  to listen for the `beforeunload` event to send this websocket message

AlejandroEsquivel (changes requested):

covalent_ui/webapp/src/components/terminal/Terminal.js
Prasy12: Double checked and it seemed to use the same socket connection on re-renders as well.

covalent_ui/api/main.py
```diff
@@ -65,6 +73,62 @@ async def chat_message(sid, data):
...
+
+async def read_and_forward_pty_output():
+    max_read_bytes = 1024 * 10
+    while terminal_subprocess:
+        await sio.sleep(0.01)
+        if file_descriptor and terminal_subprocess:
+            timeout_sec = 0
+            (data_ready, _, _) = select.select([file_descriptor], [], [], timeout_sec)
+            if data_ready:
+                output = os.read(file_descriptor, max_read_bytes).decode()
```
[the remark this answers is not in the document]
Aravind-psiog: @AlejandroEsquivel file_descriptor is created on pty and since it is an integer value, It cannot be read from aiofiles as a file.

Aravind-psiog:
@AlejandroEsquivel We have pushed the fixes for the above comments. Can you review and help merge it with develop? Please let us know if everything is good.

covalent_ui/webapp/src/components/terminal/Terminal.js
Aravind-psiog: Great catch @AlejandroEsquivel , we have now included the stop_terminal logic inside the disconnect socket function which will take care of this particular case and connect back to the UI without any problem.

covalent_ui/api/main.py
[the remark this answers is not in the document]
AlejandroEsquivel: Got it @Aravind-psiog, this should suffice for now

covalent_ui/webapp/src/components/terminal/Terminal.js
AlejandroEsquivel: Thanks @Aravind-psiog that seems like a good solution

AlejandroEsquivel (approved):

PR #1332, opened 2022-10-12
Handle exceptions when instantiating an executor

### LINKED ISSUE #1331: Errors when instantiating executors can cause workflows to hang
When `_run_task` encounters an error when instantiating an executor, the workflow can hang. This occurs for the following reason.

There are [two sources of failure](https://github.com/AgnostiqHQ/covalent/blob/b45fef20b109607173752f447b8e7ed761620338/covalent_dispatcher/_core/execution.py#L246) in `_run_task`: `Executor.execute()`, and instantiating executors. Only the former is handled correctly.

When electrons raise exceptions in `_run_task()`, the exception is caught by `_run_task` and returned in the node result. The failed node result triggers a [handler](https://github.com/AgnostiqHQ/covalent/blob/develop/covalent_dispatcher/_core/execution.py#L421) that "poisons" the task queue to [signal the dispatcher](https://github.com/AgnostiqHQ/covalent/blob/develop/covalent_dispatcher/_core/execution.py#L626) to stop submitting tasks and clean up.

Unfortunately, exceptions from instantiating executors are simply [re-raised](https://github.com/AgnostiqHQ/covalent/blob/b45fef20b109607173752f447b8e7ed761620338/covalent_dispatcher/_core/execution.py#L288). This behavior is incorrect because it bypasses [all node handlers](https://github.com/AgnostiqHQ/covalent/blob/develop/covalent_dispatcher/_core/execution.py#L463) and will cause the dispatcher to listen on the `tasks_queue` indefinitely if additional tasks remain to be submitted.

The solution is to always return a `node_result` from `_run_task`; all exceptions must be caught and handled.

cjao:

Closes #1331.


cjao:
@AlejandroEsquivel why was codecov skipped?

AlejandroEsquivel (approved):
Awesome thanks for looking into this @cjao

Issue #1260, opened 2022-09-20
Show a helpful message if trying to dispatch when Covalent is not started


scottwn:
### Description

If you attempt to dispatch when Covalent is not started, you get a trace with a `ConnectionRefusedError: [Errno 61] Connection refused`. For an experienced user, it's obvious this is caused by the dispatcher server not running, but for others it might not be so obvious. We should implement more graceful exception handling, so that the user gets a helpful message like "The Covalent dispatcher server is not running. You must start Covalent before dispatching your workflow."

### Acceptance Criteria
- [x] The `LocalDispatcher.dispatch` method displays a pretty message describing the error.
- [x] The error message appears when a `ConnectionRefusedError` is raised.
- [x] The error message is printed using Covalent's logger, `logger.app_log`, and is raised at the `ERROR` log level.
- [x] A unit test is created in the corresponding test file, `tests/covalent_tests/dispatcher_plugins/local_test.py` (this file does not currently exist). Note that unit tests in the Covalent CI are run _without_ the server running, so any attempt to dispatch a workflow within the unit test should return this error message.

rutvikk05:
Hello, I would like to work on this issue. Could you please assign it to me?

wjcunningham7:
Thanks for participating in Hacktoberfest @rutvikk05!  Looking forward to your contribution :+1:

rutvikk05:
Thanks @wjcunningham7!
I've started working on this, but not sure how to approach the problem. In which file should I make the changes to change the error message?

wjcunningham7:
@rutvikk05 the file you'll want to take a look at is `covalent/_dispatcher_plugins/local.py`.  That will include the classes and methods referenced in the acceptance criteria.

rutvikk05:
@wjcunningham7 I'm trying to reproduce the error in windows 11, but I'm unable to setup covalent. Is it compatible with windows?

scottwn:
Windows is not supported. [Here is the list of supported platforms](https://covalent.readthedocs.io/en/latest/getting_started/compatibility.html).

scottwn:
hey @Aravind-psiog, I no longer have responsibility for issue assignment, @wjcunningham7 can assign the issue

Aravind-psiog:
PR created @scottwn @wjcunningham7 . https://github.com/AgnostiqHQ/covalent/pull/1364. Please let me know if anything has to be done.

PR #1364, opened 2022-10-18
fix for #1260

Aravind-psiog:


covalent/_dispatcher_plugins/local.py
```diff
@@ -82,10 +85,14 @@ def wrapper(*args, **kwargs) -> str:
             json_lattice = lattice.serialize_to_json()
 
             test_url = f"http://{dispatcher_addr}/api/submit"
-
-            r = requests.post(test_url, data=json_lattice)
-            r.raise_for_status()
-            return r.content.decode("utf-8").strip().replace('"', "")
+            try:
+                r = requests.post(test_url, data=json_lattice)
```
venkatBala, line 89: should we call this something else other than `test_url`? @AlejandroEsquivel @cjao @kessler-frost

venkatBala (changes requested):

covalent/_dispatcher_plugins/local.py
Aravind-psiog: Could we change it to dispatcher_url?

tests/covalent_tests/dispatcher_plugins/local_test.py
```diff
@@ -0,0 +1,47 @@
...
+    @ct.lattice
+    def simple_workflow(a, b):
+        phrase = join_words(a, b)
+        return excitement(phrase)
+
+    # Dispatch the workflow
+    dispatch_id = ct.dispatch(simple_workflow)("Hello", "World")
+    assert (
+        dispatch_id
+        == 'Covalent server is not running. Start the server using "covalent start" and try dispatching workflows.'
```
FyzHsn, line 45: I don't know if it should be returning the dispatch id set to a string explaining the server is not running. Seems like it would be more appropriate to raise an error.

FyzHsn (changes requested):
Left a blocking comment for this PR. Also, could you add unit tests covering the lines of code added? The PR can't be merged if the diff coverage isn't greater than 100%. Thanks.

tests/covalent_tests/dispatcher_plugins/local_test.py
Aravind-psiog: @FyzHsn I have made the changes regarding the string set which is assigned to the dispatch id.

Aravind-psiog:
> Left a blocking comment for this PR. Also, could you add unit tests covering the lines of code added? The PR can't be merged if the diff coverage isn't greater than 100%. Thanks.

unit test failed to cover 100% because the `tests/covalent_tests/dispatcher_plugins/local_test.py` runs after the covalent server is started in the git actions. So I had written this test to pass even if the server was running and thought of letting you know afterward. If it is ok can I push the modified test case? However, it will fail until we change the git actions.

tests/covalent_tests/dispatcher_plugins/local_test.py
AlejandroEsquivel: @Aravind-psiog I agree with @FyzHsn such that there should be an error raised when calling `ct.dispatch` when the server is running. Currently there is still a string being returned when it should be an exception raised.

AlejandroEsquivel (changes requested):

Aravind-psiog:
> > Left a blocking comment for this PR. Also, could you add unit tests covering the lines of code added? The PR can't be merged if the diff coverage isn't greater than 100%. Thanks.
> 
> unit test failed to cover 100% because the `tests/covalent_tests/dispatcher_plugins/local_test.py` runs after the covalent server is started in the git actions. So I had written this test to pass even if the server was running and thought of letting you know afterward. If it is ok can I push the modified test case? However, it will fail until we change the git actions.

@AlejandroEsquivel I have the changes that @FyzHsn suggested. Regarding the test coverage, as I mentioned I have written a test case to pass when a workflow is dispatched when the server is not running. As expected, the server didn't start in the git actions and the tests passed but codecov failed to meet its requirement. The reason behind this is it did not cover the code block which will execute only when the server is started. So is there a way to cover this code block without running the server?

Thanks

madhur-tandon:
@kessler-frost and now this can be closed since you merged mine!

madhur-tandon:
Tagging @santoshkumarradha here as well.

PR #1377, opened 2022-10-24
Add folder upload/download to/from s3 buckets

### LINKED ISSUE #1316: File Transfer not working for Folders
### Environment

- **Covalent version**: 0.199.0rc0
- **Python version**: Python 3.8.13
- **Operating system**: macOS Monterey (v 12.5.1)

### What is happening?

I encountered the error **(404) when calling the HeadObject operation: Not Found** when attempting to transfer a folder from an S3 bucket to my local directory. Similar syntax does work with individual files, so this is an issue with transferring folders only.

I defined the file transfer as below:

```
strategy = ct.fs_strategies.S3(credentials="~/.aws/credentials", profile="default", region_name="us-east-1")
from_remote_file = ct.fs.Folder("s3://my_data/folder", is_remote=True)
to_local_file = ct.fs.Folder("/My_local_dir/")
ft = ct.fs.FileTransfer(from_remote_file, to_local_file, strategy=strategy, order=ct.fs.Order.BEFORE)
```

And called it in the electron like this:

```
@ct.electron(files=[ft])
def preprocess_data(files=[]):
    y = "test"
    return y
```

with the lattice: 

```
@ct.lattice
def dummy_workflow(input):
    y = preprocess_data()
    return y
```

### What should happen?

The folder should be transferred from the S3 bucket to the specified local directory when the workflow is run.

### Any suggestions?

_No response_

FyzHsn:
Fixes #1316 


AlejandroEsquivel (approved):
Looks great thanks @FyzHsn

PR #1386, opened 2022-10-25
Clarify CLI message when db needs to be upgraded

### LINKED ISSUE #1385: Clarify CLI output when DB needs to be migrated
When users start Covalent for the first time or after a DB schema update, the following message is printed to the console:
```
There have been changes to the database.
    (use "covalent db migrate" to run database migrations)
```
It is not obvious from this that the dispatcher has not actually started. The message should clarify that users need to re-try `covalent start` after the db migration, something like
```
Covalent not started. The database needs to be upgraded.
    (use "covalent db migrate" to run database migrations and retry "covalent start")
```

cjao:

Closes #1385.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Enabling logging by default
 - Removed debugging output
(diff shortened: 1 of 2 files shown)


AlejandroEsquivel (approved):
Thanks for this @cjao 🥇

PR #1391, opened 2022-10-27
Improve handling of electron errors during runtime

### LINKED ISSUE #1373: Improvement in infrastructure teardown in executors
### What should we add?

In `AsyncBaseExecutor` for example, we do the following calls in this order: `setup`, `run`, and `teardown`, which are essentially responsible for setting up the infrastructure, running the task function in that infrastructure and then tearing that infra down respectively.

But we are not taking care of the case where `run` step might fail - which will mean `teardown` is never called. This means for any failed task the infra will still be up and running without notifying the user at all. Hence, the proposal is to put the `run` step in a `try` and `teardown` in a finally block so that the teardown always happens if the `setup` step was successful.

Essentially this might be something that we want happening, in order to prevent unnecessary costs to the user:
```python
await self.setup(task_metadata=task_metadata)
try:
	result = await self.run(function, args, kwargs, task_metadata)
finally:
	await self.teardown(task_metadata=task_metadata)
```

The tests for this should make sure the infra is teared down after a failed task. A similar change might also be required for the `BaseExecutor` to support this for the non-async aware executors.

_No response_

### LINKED ISSUE #1380: Stdout and stderr from tasks are not routed to the correct graph nodes
The method Covalent currently uses to retrieve the stdout and stderr logs of each task is unreliable. When multiple tasks are run concurrently, the messages output by one task may be inadvertently attributed to another task and stored in the wrong graph node.

The following workflow reproduces the problem on my machine.
```python
from covalent.executor import DaskExecutor
from dask.distributed import LocalCluster
lc = LocalCluster()
dask_exec = DaskExecutor(lc.scheduler_address)

@ct.electron(executor=dask_exec)
def task_0():
    print("Hello from Task 0")
    return 0
    
@ct.electron(executor=dask_exec)
def task_1():
    import time
    time.sleep(2)    
    print("Hello from Task 1")
    return 1    
    
@ct.lattice
def workflow():
    task_0()
    task_1()
```
After dispatching this workflow, I retrieved the result object and inspected the `stdout` property of each graph node. Here is what I get:
```python

print("Task 0 stdout:", res.lattice.transport_graph.get_node_value(0, "stdout"))
print("Task 1 stdout:", res.lattice.transport_graph.get_node_value(1, "stdout"))
```
```
Task 0 stdout: 
Task 1 stdout: Hello from Task 0
```

If I interchange the sleep statements, the output of each task is correctly retrieved and persisted:
```python
@ct.electron(executor=dask_exec)
def task_0():
    print("Hello from Task 0")
    return 0
    
@ct.electron(executor=dask_exec)
def task_1():
    import time
    time.sleep(2)    
    print("Hello from Task 1")
    return 1

print("Task 0 stdout:", res.lattice.transport_graph.get_node_value(0, "stdout"))
print("Task 1 stdout:", res.lattice.transport_graph.get_node_value(1, "stdout"))
```
```
Task 0 stdout: Hello from Task 0

Task 1 stdout: Hello from Task 1
```

### Why this happens

Each executor's implementation of `run()` is currently expected to retrieve the `stdout` and `stderr` from the executor backend after a task completes and print those strings to `sys.stdout` and `sys.stderr`, respectively. These streams are [redirected by a context manager](https://github.com/AgnostiqHQ/covalent/blob/940ff9d986a3cdf36a0e8f2b435056aa416ff1b6/covalent/executor/base.py#L385) in the base executor's implementation of `execute()` and [returned to the dispatcher](https://github.com/AgnostiqHQ/covalent/blob/940ff9d986a3cdf36a0e8f2b435056aa416ff1b6/covalent/executor/base.py#L397). For example, [here is how the Dask executor captures the stdout for a task](https://github.com/AgnostiqHQ/covalent/blob/940ff9d986a3cdf36a0e8f2b435056aa416ff1b6/covalent/executor/executor_plugins/dask.py#L114).

Since all executor instances monitor the same `sys.stdout` file descriptor, this technique breaks down when multiple tasks are writing to that file descriptor. The context manager for one task could inadvertently capture the output printed by the `run()` method for another task.

It would seem better for each task to print to its own "stdout" and "stderr" file descriptors which are not shared with any other task. 

### Design

**Acceptance Criteria**
- For `BaseExecutor` and `AsyncBaseExecutor`
    - [ ] The redirection logic for `sys.stdout` and `sys.stderr` is removed.
    - [ ] Task-specific streams `self._task_stdout` and `self._task_stderr` are instantiated at the beginning of `execute()`. These attributes should be defined only at the beginning of `execute()` **and not in the executor constructor** so that they don't appear during workflow construction. They should be made accessible to `run()` as the properties `self.task_stdout` and `self.task_stderr`, respectively.

-  Local and Dask executors should be adjusted:
    - [ ] The `stdout` and `stderr` generated by the task are printed to `self.task_stdout` and `self.task_stderr`, respectively.
    - [ ] When there are concurrent tasks, the `stdout` and `stderr` for each task is stored in the correct transport graph node; the test cases in [#1380](https://github.com/AgnostiqHQ/covalent/issues/1380) should pass.

- [ ] Add a functional test involving a workflow with multiple electrons concurrently printing to stdout and stderr. Verify that the messages generated by each task are stored in the correct transport graph node.

### LINKED ISSUE #1390: Improve how Covalent retrieves and reports electron errors
### Background
Users may encounter several conceptually distinct classes of irrecoverable errors while using Covalent:

- Task runtime exceptions that occur in an executor backend while running a task. These failures may reflect oversights in the workflow's business logic and would normally be printed to stderr if a workflow were run locally rather than dispatched to Covalent.

- Task-specific exceptions from Covalent itself, resulting from an abnormal interaction between the Covalent dispatcher and an executor backend, such expired AWS credentials or a temporary loss of network connectivity.

- Errors pertaining to an overall workflow and not any specific task, such as failure in postprocessing.

- Server errors not specific to a workflow, for instance, resource starvation or loss of connection to the database.

Covalent does store all of these types of errors. However, it is not easy for users to retrieve or distinguish them.

There is also a design defect in how each task's `stdout` and `stderr` are returned to the dispatcher. As explained in Covalent issue [#1380](https://github.com/AgnostiqHQ/covalent/issues/1380), when multiple tasks run concurrently, **the log output for one task may be inadvertently attributed to another task**.  The proposed remedy for this is discussed in that issue.

**Main objectives for this issue**:
* Define clear rules for how Covalent handles different types of errors.
* Ensure that stdout and stderr are recovered when task runtime exceptions are raised.
* Make all task-level exceptions visible in the UI.

### Design

**Corresponding to each of the above classes of irrecoverable errors, we will implement the following error handling rules:**

* Task runtime errors are appended to the task's `stderr`. The task is marked as failed, but any existing stdout and stderr are returned to the dispatcher. This mimics the behavior when a program is run directly instead of being dispatched; the exception traceback is typically printed at the end of other program output.
* The node `error` attribute is reserved for task-specific errors originating in Covalent itself. These may pertain to the interaction between Covalent and the executor backend, or in some cases Covalent may have trouble instantiating the executor specified for a task. 

The two types of errors are mutually exclusive from Covalent's perspective; if the dispatcher cannot communicate normally with the executor backend, it cannot extract any task output.

**When task-specific errors occur, the lattice-level `Result.error` will summarize the failure by enumerating the `node_id` and name for each failed task.**

- [ ] Workflow level exceptions (currently postprocessing errors) are saved in `Result.error`.

**Acceptance Criteria**
- [ ] DB schema is updated to v13:
    - [ ] Remove `Electron.info_filename`, add `Electron.error_filename` in `models.py`
    - [ ] Copy these changes to `data_layer/electron_dal.py` and `api/v1/database/schema/electron.py` in `covalent_ui`
    - [ ] Adjust `upsert.py` and `write_result_to_db.py` accordingly.
    - [ ] Generate alembic migration script to v13.
    

- [ ] The UI displays errors for each failed electron.
    - [ ] When a task raises a exception, its `stderr` and the exception traceback are **both** displayed in the Electron ErrorCard (on the right panel when one clicks on a failed task).
    - [ ] When a task fails due to an exception raised by Covalent, the exception is displayed in the Electron ErrorCard. 
    
This can be accomplished by modifying `covalent_ui/api/v1/routes/end_points/electron_routes.py`: When the `"error"` detail [is requested of an electron](https://github.com/AgnostiqHQ/covalent/blob/15089219bdb609b6b5149fbe1aa520b98dc1b52d/covalent_ui/api/v1/routes/end_points/electron_routes.py#L156), the response shall concatenate both the electron's `stderr` and `error`.

- [ ] Define `TaskRuntimeError` exception in `covalent/_shared_files/exceptions.py`. This is to be raised by each executor plugin's implementation of `run()` if the task experiences an unhandled exception in the executor backend.

-  [ ] Base executors are modified in `base.py`: (Async)BaseExecutor
    - [ ] `execute()` returns a 4-tuple `(output, stdout, stderr, exception_raised)` of type `(Any, str, str, bool)`, where:
        - `exception_raised` indicates whether a `TaskRuntimeError` was raised in `run()`.
        - `stdout` and `stderr` are the messages generated by the task in the executor backend.
        - `output` is either the task's return value, if the task completed successfully, or `None` if the task raised an unhandled exception.

- [ ] Local and Dask executors are adjusted:
    - [ ] When a task raises an exception, the full traceback is appended to the `stderr` already generated by the task. Any stdout and stderr that has already been printed should be recovered. **The final act of `run()` in this case should be to raise `TaskRuntimeError()` to notify `execute()` that an error occurred in the executor backend**.
    
- [ ] `execution.py` reflects the new error handling methods:
    - If `Executor.execute()` returns control to `_run_task` with `exception_raised=True`, the task is marked as failed, but stdout and stderr are saved to the node result. The node `error` attribute remains empty.
    - For any other exception occurring in `_run_task`, the exception is be handled, the task is marked as failed, and the node `error` attribute is populated with either
        - the full traceback if Covalent is started with the "debug" log level
        - just the string representation of the exception otherwise. This shall be the default behavior to avoid polluting the error message with irrelevant Covalent internals.
    - Whenever a workflow fails, the `Result._error` attribute either:
        - summarizes the node id and task name of each failed task, or
        - describes the failure if the reason for workflow failure is not specific to a task (e.g. post processing failed)
    - The `Result.status` property is set to `Result.FAILED` whenever `_run_planned_workflow` raises an unhandled exception.

- [ ] Unit tests achieve 100% patch coverage.

- [ ] Add basic functional tests to check reporting of both task runtime errors (task raises an exception) and task-specific errors raised by Covalent (e.g. bad executor). 
    - [ ] The tests should verify that if tasks print something to `sys.stdout` and `sys.stderr` before raising an exception, those streams are recovered by Covalent and stored in each node's `stdout` and `stderr` property. In addition, they should verify that exceptions raised by Covalent are saved to the node's `error` attribute.
    - [ ] The tests shall include concurrently running tasks and verify that the stdout and stderr for each task is stored in the correct transport graph node.

cjao:

Closes #1390.
Closes #1380.
Closes #1373.


covalent_dispatcher/_db/models.py
```diff
@@ -182,9 +182,8 @@ class Electron(Base):
     # Name of the file containing standard error generated by the task
     stderr_filename = Column(Text)
 
-    # Name of the file containing execution information generated at
-    # runtime (such as Slurm logs)
-    info_filename = Column(Text)
```
AlejandroEsquivel, line 187: @cjao is this just a renaming of info_filename to error_filename ? do they contain effectively the same content?

covalent/executor/utils/local_executor_utils.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+
+import io
+import multiprocessing as mp
+import traceback
+from contextlib import redirect_stderr, redirect_stdout
+from typing import Callable, Dict, List
+
+
+def local_wrapper(fn: Callable, args: List, kwargs: Dict, q: mp.Queue):
```
AlejandroEsquivel, line 32: I see there is code duplication here with `dask_wrapper` @cjao 

Does it maybe make sense for the AbtsractBaseExecutor to have a method that is in general mediating the actual execution on a function with the redirect stream logic and properly formatting the traceback? it may be possible more executors such as the Executor or AsyncCustomExecutor to potentially re-use this method

AlejandroEsquivel (commented):
Looks great @cjao , love how this is also reverse compatible with the aws plugins as well. Some minor comments, let me know your thoughts

covalent/executor/utils/local_executor_utils.py
cjao: Thanks for your feedback! This is actually subtly different from the `dask_wrapper` because of the multiprocessing Queue parameter. In effect, `local_wrapper` replicates the logic used by [ProcessPoolExecutor](https://github.com/python/cpython/blob/3.11/Lib/concurrent/futures/process.py#L255-L260) in concurrent.futures.

covalent_dispatcher/_db/models.py
cjao: The `info_filename` was never actually used, and the semantics are different. Here, `error_filename` contains errors generated by `Covalent` during a task's execution, not by the executor backend.

AlejandroEsquivel: Sounds good @cjao , if it was indeed used I would have asked to include a data migration as well to update the `error_filename` with the `info_filename` values but all good here!

covalent/executor/utils/local_executor_utils.py
AlejandroEsquivel: Got it @cjao , given that these wrapper functions are still tied with a particular executor it may make sense to make them class methods of the particular executor they are tied to as they seem to share the same logical grouping. As for lines 35-41 however those are the lines which are repeated for dask/local and might be sensible to expect that those could also be used by other executors so may have sense to factor that into it's own function where the dask/local wrapper functions utilize that new method. Let me know your thoughts on this, thanks!

cjao: This is a great suggestion. The wrappers used by local and dask are stored in a wrappers module under `covalent/executors/utils`.

AlejandroEsquivel (approved):
Looks good @cjao

PR #1392, opened 2022-10-27
Ensure that each node output is a TransportableObject when printing

### LINKED ISSUE #1393: Result object cannot be printed if a node fails
Consider the following trivial workflow:

```python
@ct.electron(executor="local")
def task():
    raise RuntimeError("error")

@ct.lattice(executor="local")
def workflow():
    return task()
```
I cannot print the result object:
```
result_object = ct.get_result(dispatch_id, wait=True)
print(result_object)
```

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [18], in <cell line: 11>()
      9 dispatch_id = ct.dispatch(workflow)()
     10 result_object = ct.get_result(dispatch_id, wait=True)
---> 11 print(result_object)

File ~/Agnostiq/code/covalent/covalent/_results_manager/result.py:127, in Result.__str__(self)
    125 node_outputs = self.get_all_node_outputs()
    126 for k, v in node_outputs.items():
--> 127     show_result_str += f"{k}: {v.object_string}\n"
    129 return show_result_str

AttributeError: 'NoneType' object has no attribute 'object_string'
```

This happens because when Covalent encounters an exception when processing a task, the transport graph node's `output` property is set to `None`; for any task that completes normally, the output is a `TransportableObject` (tasks that don't return a value have output `TransportableObject(None)`).

A simple fix is to convert all `None` node outputs to `TransportableObject(None)` in the `Result.__str__` method; the idempotent `TransportableObject.make_transportable()` method is convenient for this.

cjao:
Convert None outputs to TransportableObject(None)

Closes #1393.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for transferring the contents of folders to and from S3 buckets using the file transfer module.
 
+### Fixed
+
+- Result objects now print correctly when nodes fail
+
 ## [0.207.0-rc.0] - 2022-10-26
 
 ### Authors
--- a/covalent/_results_manager/result.py
+++ b/covalent/_results_manager/result.py
@@ -124,7 +124,7 @@ Node Outputs
 
         node_outputs = self.get_all_node_outputs()
         for k, v in node_outputs.items():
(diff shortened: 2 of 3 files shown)


AlejandroEsquivel (approved):
@cjao for the win!  🙌🏼

PR #1394, opened 2022-10-28
Fix for 1284

### LINKED ISSUE #1284: Electrons with unspecified executors do not inherit the Lattice-level executor
### Description

Electrons that don't explicitly specify an executor in their decorator should inherit the executor provided in the Lattice decorator. If no executor is provided in the lattice decorator, the electron should use the Covalent default executor ("local" or "dask" depending on whether Covalent is started with `--no-cluster`).

However, this executor inheritance doesn't actually work. Consider the following example:

```python
@ct.electron
def task(x):
    return x

@ct.lattice(executor="awslambda")
def workflow(x):
    return task(x)

workflow.build_graph(1)
```

Since the `task` electron doesn't specify an executor, we expect the Lattice-level executor `awslambda` to apply to that electron. This is not what happens:

```
print(workflow.metadata["executor"])
print(workflow.transport_graph.get_node_value(0, "metadata")["executor"])
```

```
awslambda
local
```

The same problem occurs for other metadata such as `call_before` or `deps`; however, to minimize scope this issue only tracks the propagation of executors from lattices to their electrons. Other metadata can be dealt with similarly in followup issues.

Here are some places to start your debugging:
- The [Electron decorator](https://github.com/AgnostiqHQ/covalent/blob/32929263ab5defc3232f72b8da2c2fe8facf33f9/covalent/_workflow/electron.py#L519) (implements `@ct.electron`)
- [The place where electrons are supposed to inherit lattice-level metadata](https://github.com/AgnostiqHQ/covalent/blob/32929263ab5defc3232f72b8da2c2fe8facf33f9/covalent/_workflow/electron.py#L322).

### Acceptance criteria

- [ ] The Lattice-level executor shall propagate to any electron whose executor is not explicitly set.
- [ ] Any explicit executor setting for an electron shall take precedence over the lattice's setting.
- [ ] Add unit tests checking both of the above two scenarios.

RaviPsiog:

Closes #1284.


tests/covalent_tests/executor/lattice_to_electrons.py
```diff
@@ -0,0 +1,96 @@
...
+    hello_world = LatticeClass.deserialize_from_json(hello_world.serialize_to_json())
+
+    for i in data["nodes"]:
+        if ("parameter" not in i["name"]) and ("excitement" in i["name"]):
+            assert i["metadata"]["executor"] == "local"
+        elif "parameter" not in i["name"]:
+            assert i["metadata"]["executor"] == hello_world.metadata["executor"]
+
+
+test_electrons_have_lattice_executor()
```
FyzHsn, line 95: Remove test method calls.

tests/covalent_tests/executor/lattice_to_electrons.py
```diff
@@ -0,0 +1,96 @@
...
+    data = json.loads(data)
+    hello_world = LatticeClass.deserialize_from_json(hello_world.serialize_to_json())
+
+    for i in data["nodes"]:
+        if "parameter" not in i["name"]:
+            # print(i["name"], i["metadata"]["executor"])
+            assert i["metadata"]["executor"] == hello_world.metadata["executor"]
+
+
+# if an electron is changed to something else, then it should be taken instead of lattice executor.
```
FyzHsn, line 62: Move comment to inside a doc string

tests/covalent_tests/executor/lattice_to_electrons.py
```diff
@@ -0,0 +1,96 @@
...
+
+"""Unit tests to check that electrons inherit the parent lattice executor."""
+
+import json
+
+import covalent as ct
+from covalent._workflow.lattice import Lattice as LatticeClass
+
+
+# start the covalent server with no cluster and then run this program,
```
FyzHsn, line 29: Move comment to inside doc string.

covalent/_workflow/electron.py
```diff
@@ -516,13 +525,27 @@ def wait_for(self, electrons: Union["Electron", Iterable["Electron"]]):
...
+
+
 def electron(
     _func: Optional[Callable] = None,
     *,
     backend: Optional[str] = None,
-    executor: Optional[
-        Union[List[Union[str, "BaseExecutor"]], Union[str, "BaseExecutor"]]
-    ] = DEFAULT_METADATA_VALUES["executor"],
+    executor: Optional[Union[List[Union[str, "BaseExecutor"]], Union[str, "BaseExecutor"]]] = "",
```
FyzHsn, line 548: Are you sure `DEFAULT_METADATA_VALUES["executor"]` should be removed?

covalent/_workflow/electron.py
```diff
@@ -516,13 +525,27 @@ def wait_for(self, electrons: Union["Electron", Iterable["Electron"]]):
         )
 
 
+# def electron(
```
FyzHsn, line 528: Remove commented out lines of code.

FyzHsn (changes requested):
Requested some changes. Also, could you add unit tests in addition to the integration tests added below?

covalent/_workflow/electron.py
```diff
@@ -322,12 +322,21 @@ def __call__(self, *args, **kwargs) -> Union[Any, "Electron"]:
...
+                    k not in consumable_constraints
+                    and k in DEFAULT_METADATA_VALUES
+                    and self.get_metadata(k) is DEFAULT_METADATA_VALUES[k]
+                ):
+                    self.set_metadata(k, active_lattice.get_metadata(k))
+            # if electron executor metadata is not empty string, then it should assigned with whatever it has defined.
+            # else, we need to set the electron executor to be of lattices', which will be of default.
+            # this case is not possible without having an empty executor string at first, so i had to change this also at line 550
+            else:
+                if self.get_metadata(k) != "":
```
RaviPsiog, line 336: Hi @FyzHsn  For the question of whether am i sure to remove `DEFAULT_METADATA_VALUES["executor"]` .
yes, will explain what I faced. 
First I coded line 336 as `if self.get_metadata(k) != DEFAULT_METADATA_VALUES["executor"]`
And that worked fine for all cases, except for one case. That goes as follows: 
Consider covalent started with dask, And lattice executor changed to local as `@ct.lattice(executor="local")`. Now all the electrons are supposed to get local. Then I changed one electron executor to default(dask) as `@ct.electron(executor="dask")`. Now as electrons precede lattice, it is supposed to take dask. But it doesn't, according to line 336 I coded at first, as the newly changed executor is actually the default one, and that makes it fail the if condition and gets local instead!! This can be avoided if I have the default electron executor as an empty string, **then we can accomplish the case I mentioned, where we change the electron executor to default one** as per the new code `if self.get_metadata(k) != ""` . so now, even if we change the electron executor to default( here - dask), it will be changed. Considering this case, I had to change the default executor to empty string. Please let me know, if the explanation is not clear.

RaviPsiog:
> Requested some changes. Also, could you add unit tests in addition to the integration tests added below?

Sure, Noted. I will make the changes and will let you know @FyzHsn

RaviPsiog:
Hi @FyzHsn , Above mentioned changes were made and wrote the unit test.

RaviPsiog:
Hi @FyzHsn , The unit test that I wrote is supposed to cover line 330.  Then I saw that the line is not covered as we are using the "is" operator, which checks the reference. If i am not wrong, checking if they are identical is enough in this case. And hence i changed it to == operator. Now, The coverage should meet the requirements and also it's working as it's supposed.

FyzHsn (dismissed):

FyzHsn (approved):

PR #1395, opened 2022-10-28
Fix for 1291

### LINKED ISSUE #1291: Sublattices without explicitly specified executors don't inherit the parent lattice's executor.
### Description

When users do not specify an executor in a lattice decorator, the lattice's executor is intended to be automatically set as follows:
- If the lattice is top-level, use the default executor from the Covalent config.
- If the lattice is a sublattice, use its parent lattice's executor.

Consider the following simple lattice consisting of a single sublattice. The top-level lattice's executor is set to `"dask"`, whereas the sublattice's executor is unspecified.
```python
@ct.electron
def task(x):
    return x*x

@ct.electron
@ct.lattice
def subworkflow(x):
    return task(x)

@ct.lattice(executor="dask")
def workflow(x):
    return subworkflow(x)

dispatch_id = ct.dispatch(workflow)(5)
result_object = ct.get_result(dispatch_id, wait=True)
sub_result_object = result_object.lattice.transport_graph.get_node_value(0, "sublattice_result")
```

Since the sublattice does not specify an executor, it should inherit its parent lattice's executor. However, this is not what happens; the sublattice's executor is set to `"local"`, the default executor:

```python
print(result_object.lattice.metadata["executor"])
print(sub_result_object.lattice.metadata["executor"])
```

```
dask
local
```

### Acceptance criteria

- [ ] If a sublattice does not specify an executor, it should inherit its parent lattice's executor.
- [ ] Explicit executor settings for a sublattice shall be honored.
- [ ] If a top-level lattice (not a sublattice) does not specify an executor, it should use the Covalent default executor -- `"local"` if Covalent is started with the `--no-cluster` option and `"dask"` otherwise.
- [ ] Add tests to verify executor settings for lattices and sublattices in all three of the above scenarios.

Some suggestions to start debugging:
- The [lattice decorator](https://github.com/AgnostiqHQ/covalent/blob/develop/covalent/_workflow/lattice.py#L307)
- The server's [internal logic for running sublattices](https://github.com/AgnostiqHQ/covalent/blob/develop/covalent_dispatcher/_core/execution.py#L191); see also `_build_sublattice_graph`.
- The related issue #1284 for automatically setting electron executors.

RaviPsiog:
Closes #1291.


FyzHsn (changes requested):
I've suggested some changes that need to be made in order for this PR to get merged. Furthermore, 

1. The tests written here are integration tests involving multiple methods. Could you add a unit test for the method that was modified? I understand this is slightly out of scope, but moving forward, the protocol is to add method/function level unit tests at least. You can still keep the integration tests added here.
2. The test coverage has to be greater than 95% in order for this PR to be merged.

RaviPsiog:
Sure, I will write the unit tests, make the appropriate changes you mentioned, and will let you know @FyzHsn

RaviPsiog:
Hi @FyzHsn , Above mentioned changes were made and wrote the unit test. can you review it.

FyzHsn:
> Hi @FyzHsn , Above mentioned changes were made and wrote the unit test. can you review it.

Hi @RaviPsiog, the diff coverage is still at 83% and needs to be 95% before this PR can be merged. You just need to cover line 330 in `electron.py` in your tests. This PR should be good to be merged as soon as the diff coverage meets the minimum requirement

RaviPsiog:
Hi @FyzHsn , The unit test that I wrote is supposed to cover line 330.  Then I saw that the line is not covered as we are using the "is" operator, which checks the reference. If i am not wrong, checking if they are identical is enough in this case. And hence i changed it to == operator. Now, The coverage should meet the requirements and also it's working as it's supposed.

FyzHsn (approved):

PR #1416, opened 2022-11-14
Fix electron and sublattice metadata inheritance

### LINKED ISSUE #417: Sublattices should inherit the executor from the parent lattice if none is provided
A user should be able to assign an executor to a lattice and that should be used for those electrons which do not have an explicitly assigned executor.

Edit: This seems to have resolved for other cases due to changes to the codebase but still remains an issue with sublattices.

cjao:

Re-addresses #1284 and #1291.

Closes #417 (see explanation of changes [here](https://github.com/AgnostiqHQ/covalent/issues/417#issuecomment-1258529112) and in the following comment).

Fixes tests


AlejandroEsquivel (approved):
Looks great, thanks @cjao

PR #1437, opened 2022-12-05
update references to results_dir

### LINKED ISSUE #1382: Removal of `results_dir` from UX
References to the `results_dir` need to be removed from covalent. Since it might be a big change requiring modifications in a lot of files, in this issue the objective should be to remove the references to `results_dir` and its assignment etc. from the UX's side.

So, removal from the `results.py`, `electron.py`, `lattice.py`, RTDs (especially "How To"s ) and such files should be considered the acceptance criteria for resolution of this issue.

When the server constructs a `Result` object from the client's submission, its `results_dir` should be initialized to (in descending order of priority):
1) the value of the `COVALENT_DATA_DIR` env variable, if set, or
2) the value of the config key `dispatcher.results_dir`, which shall default to `~/.local/share/covalent/data`

### Acceptance Criteria
- [x] The SDK no longer makes any reference to `results_dir`.
- [x] When a workflow is submitted to Covalent and run, the `results_dir` property of the `Result` object respects the `COVALENT_DATA_DIR` environment variable if that is set. If that is unset, the config value of `dispatcher.results_dir` shall be used. **The value of `Results.results_dir` must not depend on either 1) the cwd of the Covalent dispatcher or 2) the directory from which the client submits the workflow**
- [x] The config default for `dispatcher.results_dir` is updated to `$HOME/.local/share/covalent/data`.

venkatBala:

closes #1382


cjao:
Hi @venkatBala , can you try running the dispatcher tests? The results directory is consumed there.

cjao (changes requested):
Hi @venkatBala , this is looking good. I think [execution.py](https://github.com/AgnostiqHQ/covalent/blob/35833e192d0e28d5ce1cbe9607ad7e1cf55fb1c2/covalent_dispatcher/_core/execution.py#L867) also needs to be adjusted.

AlejandroEsquivel:
@venkatBala could you address @cjao's comment when you get a chance?

cjao (approved):
Looks great @venkatBala !

PR #1554, opened 2023-03-07
Implement postprocessing electron

### LINKED ISSUE #1545: Convert post-processing to an electron
Remove postprocessing from the [runner.](https://github.com/AgnostiqHQ/covalent/blob/os-enhancements-preview/covalent_dispatcher/_core/runner.py) **All electrons look the same to the runner — no more `workflow_executor`**
Also, the SDK should bind the lattice’s `workflow_executor` to each sublattice/postprocessing electron’s `executor` during `build_graph()`.

FyzHsn:
Fixes #1545 

### DIFF
--- a/covalent/_workflow/lattice.py
+++ b/covalent/_workflow/lattice.py
@@ -38,6 +38,7 @@ from .._shared_files.utils import get_named_params, get_serialized_function_str
 from .depsbash import DepsBash
 from .depscall import DepsCall
 from .depspip import DepsPip
+from .postprocessing import Postprocessor
 from .transport import TransportableObject, _TransportGraph, encode_metadata
 
 if TYPE_CHECKING:
@@ -246,11 +247,8 @@ class Lattice:
                     )
                     raise
 
-        from .postprocessing import Postprocessor
-
-        pp = Postprocessor(lattice=self)
-
         if get_config("sdk.exhaustive_postprocess") == "true":
+            pp = Postprocessor(lattice=self)
             pp.add_exhaustive_postprocess_node(self._bound_electrons.copy())
 
         self._bound_electrons = {}  # Reset bound electrons
--- a/covalent/_workflow/lattice.py
+++ b/covalent/_workflow/lattice.py
@@ -28,11 +28,17 @@ from contextlib import redirect_stdout
 from copy import deepcopy
 from dataclasses import asdict
 from functools import wraps
-from typing import TYPE_CHECKING, Any, Callable, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
 
 from .._shared_files import logger
+from .._shared_files.config import get_config
 from .._shared_files.context_managers import active_lattice_manager
-from .._shared_files.defaults import DefaultMetadataValues
+from .._shared_files.defaults import (
+    DefaultMetadataValues,
+    postprocess_prefix,
+    prefix_separator,
+    sublattice_prefix,
+)
 from .._shared_files.utils import get_named_params, get_serialized_function_str
 from .depsbash import DepsBash
 from .depscall import DepsCall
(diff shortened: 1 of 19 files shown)


covalent/_workflow/lattice.py
```diff
@@ -87,6 +88,9 @@ def __init__(
 
         self.workflow_function = TransportableObject.make_transportable(self.workflow_function)
 
+        # Bound electrons are defined as electrons with a valid node_id, since it means they are bound to a TransportGraph.
```
kessler-frost, line 91: I see, we have a definition here as well.

covalent/_workflow/lattice.py
```diff
@@ -237,12 +239,22 @@ def build_graph(self, *args, **kwargs) -> None:
...
                 except Exception:
                     warnings.warn(
                         "Please make sure you are not manipulating an object inside the lattice."
                     )
                     raise
 
+        from .postprocessing import Postprocessor
+
+        pp = Postprocessor(lattice=self)
+
```
kessler-frost, line 252: Can you put this inside the if condition if it is only valid when the condition is true?

kessler-frost: Also, if this import is not turning out to be cyclical then maybe we can put it on the top of the file.

covalent/_workflow/postprocessing.py
```diff
@@ -0,0 +1,170 @@
...
+log_stack_info = logger.log_stack_info
+
+
+class Postprocessor:
+    def __init__(self, lattice) -> None:
+        self.lattice = lattice
+
+    def _is_postprocessable_node(self, tg: _TransportGraph, node_id: int) -> bool:
+        """Filter nodes that should be included in postprocessing."""
+        node_name = tg.get_node_value(node_id, "name")
```
kessler-frost, line 52: Missing args and return value in docstring here.

covalent/_workflow/postprocessing.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        Post processing function to be called after the lattice execution.
+        This takes care of executing statements that were not an electron
+        but were inside the lattice's function. It also replaces any calls
+        to an electron with the result of that electron execution, hence
+        preventing a local execution of electron's function.
+        Note: Here `node_outputs` is used instead of `electron_outputs`
+        since an electron can be called multiple times with possibly different
+        arguments, but every time it's called, it will be executed as a separate node.
+        Thus, output of every node is used.
```
kessler-frost, line 83: Great explanation!

covalent/_workflow/postprocessing.py
```diff
@@ -0,0 +1,170 @@
...
+            self.lattice.post_processing = False
+            return result
+
+    def _get_node_ids_from_retval(self, retval: Union["Electron", List, Dict]) -> List[int]:
+        """Get the list of electron node ids from the return value of a lattice function.
+
+        Note: 'retval' here corresponds to the return value of a lattice function.
+
+        Args:
+            retval (_type_): _description_
```
kessler-frost, line 110: Missing description.

covalent_dispatcher/_db/update.py
```diff
@@ -87,7 +88,10 @@ def _node(
 
     Returns:
         None
+
     """
+    if not node_name:
```
kessler-frost, line 93: If comparing with `None` my suggestion for the ideal way would be to do `if node_name is None`.

kessler-frost (changes requested):
Minor changes, otherwise looks good 👍

covalent/_workflow/postprocessing.py
FyzHsn: i think it was Casey's :D

kessler-frost (approved):

covalent/_workflow/postprocessing.py
cjao: Actually it was copied from `runner.py` and formerly `execution.py` -- so probably @kessler-frost 's.

FyzHsn: 😂

PR #1568, opened 2023-04-10
Reduce pion delta

FyzHsn:

### DIFF
--- a/covalent/_shared_files/defaults.py
+++ b/covalent/_shared_files/defaults.py
@@ -63,7 +63,7 @@ def get_default_sdk_config():
         ),
         "no_cluster": "true" if os.environ.get("COVALENT_DISABLE_DASK") == "1" else "false",
         "exhaustive_postprocess": "true",
-        "eager_preprocess": "false",  # Note that exhaustive and eager options are mutually exclusive.
+        "eager_postprocess": "false",  # Note that exhaustive and eager options are mutually exclusive.
     }
 
 
--- a/covalent/_shared_files/defaults.py
+++ b/covalent/_shared_files/defaults.py
@@ -63,7 +63,6 @@ def get_default_sdk_config():
         ),
         "no_cluster": "true" if os.environ.get("COVALENT_DISABLE_DASK") == "1" else "false",
         "exhaustive_postprocess": "true",
-        "reconstruct_postprocess": "false",
     }
 
 
--- a/covalent/_shared_files/defaults.py
+++ b/covalent/_shared_files/defaults.py
@@ -63,7 +63,7 @@ def get_default_sdk_config():
         ),
         "no_cluster": "true" if os.environ.get("COVALENT_DISABLE_DASK") == "1" else "false",
         "exhaustive_postprocess": "true",
-        "reconstruct_postprocess": "false",  # Note that exhaustive and eager options are mutually exclusive.
+        "reconstruct_postprocess": "false",
     }
 
 
--- a/covalent/_shared_files/defaults.py
+++ b/covalent/_shared_files/defaults.py
@@ -63,6 +63,7 @@ def get_default_sdk_config():
         ),
         "no_cluster": "true" if os.environ.get("COVALENT_DISABLE_DASK") == "1" else "false",
         "exhaustive_postprocess": "true",
+        "eager_preprocess": "false",  # Note that exhaustive and eager options are mutually exclusive.
     }
 
 
--- a/covalent/_shared_files/defaults.py
+++ b/covalent/_shared_files/defaults.py
@@ -63,7 +63,7 @@ def get_default_sdk_config():
         ),
         "no_cluster": "true" if os.environ.get("COVALENT_DISABLE_DASK") == "1" else "false",
         "exhaustive_postprocess": "true",
-        "eager_postprocess": "false",  # Note that exhaustive and eager options are mutually exclusive.
+        "reconstruct_postprocess": "false",  # Note that exhaustive and eager options are mutually exclusive.
     }
 
 
--- a/covalent/_workflow/electron.py
+++ b/covalent/_workflow/electron.py
@@ -467,7 +467,7 @@ class Electron:
                 )
 
             # For keyword arguments
-            # Filter out kwargs to be injected by call_before calldeps at execution
+            # Filter out kwargs to be injected by call_before call_deps during execution.
             call_before = self.metadata["call_before"]
             retval_keywords = {item["attributes"]["retval_keyword"]: None for item in call_before}
             for key, value in named_kwargs.items():
@@ -485,6 +485,8 @@ class Electron:
             self.function,
             metadata=self.metadata,
             node_id=self.node_id,
+            task_group_id=self.task_group_id,
+            packing_tasks=self.packing_tasks,
         )
         active_lattice._bound_electrons[self.node_id] = bound_electron
         return bound_electron
@@ -531,7 +533,12 @@ class Electron:
             def _auto_list_node(*args, **kwargs):
                 return list(args)
 
-            list_electron = Electron(function=_auto_list_node, metadata=collection_metadata)
+            list_electron = Electron(
+                function=_auto_list_node,
+                metadata=collection_metadata,
+                task_group_id=self.task_group_id,
+                packing_tasks=True,
+            )  # Group the auto-generated node with the main node.
             bound_electron = list_electron(*param_value)
             transport_graph.set_node_value(bound_electron.node_id, "name", electron_list_prefix)
             transport_graph.add_edge(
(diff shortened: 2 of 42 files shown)


covalent/_shared_files/defaults.py
```diff
@@ -63,6 +63,7 @@ def get_default_sdk_config():
         ),
         "no_cluster": "true" if os.environ.get("COVALENT_DISABLE_DASK") == "1" else "false",
         "exhaustive_postprocess": "true",
+        "reconstruct_postprocess": "false",  # Note that exhaustive and eager options are mutually exclusive.
     }
```
kessler-frost, line 67: If they are mutually exclusive then don't we just need 1 variable indicating which one to use 🤔 .

covalent/_workflow/electron.py
```diff
@@ -383,13 +434,27 @@ def __call__(self, *args, **kwargs) -> Union[Any, "Electron"]:
...
+
+        else:
+            # Default to node_id as the task_group_id.
+            self.node_id = active_lattice.transport_graph.add_node(
+                name=self.function.__name__,
+                function=self.function,
+                metadata=self.metadata.copy(),
+                function_string=get_serialized_function_str(self.function),
+            )
+            self.task_group_id = self.node_id
```
kessler-frost, line 457: It seems that the only difference in these if/else conditions is that in the else clause there is a `task_group_id`. To prevent code duplication can't we just do:
```python
....
task_group_id = self.task_group_id if self.packing_tasks else None
....
```

covalent/_workflow/electron.py
```diff
@@ -310,9 +349,21 @@ def get_item(e, key):
...
+            # Pack with main electron except for sublattices
+            get_item_electron = (
+                Electron(function=get_item, metadata=self.metadata.copy())
+                if name.startswith(sublattice_prefix)
+                else Electron(
+                    function=get_item,
+                    metadata=self.metadata.copy(),
+                    task_group_id=self.task_group_id,
+                    packing_tasks=True,
+                )
```
kessler-frost, line 363: Duplicate code.

covalent/_workflow/electron.py
```diff
@@ -298,8 +324,21 @@ def get_attr(e, attr):
...
+            name = active_lattice.transport_graph.get_node_value(self.node_id, "name")
+            get_attr_electron = (
+                Electron(function=get_attr, metadata=self.metadata.copy())
+                if name.startswith(sublattice_prefix)
+                else Electron(
+                    function=get_attr,
+                    metadata=self.metadata.copy(),
+                    task_group_id=self.task_group_id,
+                    packing_tasks=True,
+                )
```
kessler-frost, line 338: Duplicate code.

covalent/_workflow/electron.py
```diff
@@ -276,6 +289,19 @@ def get_item(e, key):
...
+                # Pack with main electron unless it is a sublattice.
+                name = active_lattice.transport_graph.get_node_value(self.node_id, "name")
+                get_item_electron = (
+                    Electron(function=get_item, metadata=iterable_metadata)
+                    if name.startswith(sublattice_prefix)
+                    else Electron(
+                        function=get_item,
+                        metadata=iterable_metadata,
+                        task_group_id=self.task_group_id,
+                        packing_tasks=True,
```
kessler-frost, line 302: Duplicate code (in the ones below as well) should be extracted into a function instead.

covalent/_workflow/lattice.py
```diff
@@ -244,8 +244,11 @@ def build_graph(self, *args, **kwargs) -> None:
                     )
                     raise
 
-        if get_config("sdk.exhaustive_postprocess") == "true":
-            pp = Postprocessor(lattice=self)
+        pp = Postprocessor(lattice=self)
+
+        if get_config("sdk.exhaustive_postprocess") != "true":
```
kessler-frost, line 249: This condition seems unnecessarily complicated, can't we just check `sdk.reconstruct_postprocess` is "true" or not?

tests/covalent_tests/workflow/electron_test.py
```diff
@@ -349,3 +349,150 @@ def sublattice(x):
...
+    point_electron_gid = tg.get_node_value(0, "task_group_id")
+    getitem_x_gid = tg.get_node_value(1, "task_group_id")
+    getitem_y_gid = tg.get_node_value(3, "task_group_id")
+    assert point_electron_gid == 0
+    assert getitem_x_gid == point_electron_gid
+    assert getitem_y_gid == point_electron_gid
+    assert tg.get_node_value(2, "task_group_id") == 2
+    assert tg.get_node_value(4, "task_group_id") == 4
+    assert tg.get_node_value(5, "task_group_id") == 5
+    assert tg.get_node_value(6, "task_group_id") == 6
```
kessler-frost, line 417: Can do this with `assert all()` and list comprehension.

tkdtaylor (commented):
Code review looks good, recommend a 2nd set of eyes that are more familiar with the codebase to check as well though.

tests/covalent_tests/workflow/electron_test.py
```diff
@@ -349,3 +349,150 @@ def sublattice(x):
...
+    arr_electron_gid = tg.get_node_value(0, "task_group_id")
+    getitem_x_gid = tg.get_node_value(1, "task_group_id")
+    getitem_y_gid = tg.get_node_value(3, "task_group_id")
+    assert arr_electron_gid == 0
+    assert getitem_x_gid == arr_electron_gid
+    assert getitem_y_gid == arr_electron_gid
+    assert tg.get_node_value(2, "task_group_id") == 2
+    assert tg.get_node_value(4, "task_group_id") == 4
+    assert tg.get_node_value(5, "task_group_id") == 5
+    assert tg.get_node_value(6, "task_group_id") == 6
```
kessler-frost, line 457: Same as previous `assert all()` comment.

tests/covalent_tests/workflow/electron_test.py
```diff
@@ -349,3 +349,150 @@ def sublattice(x):
...
+    tup_electron_gid = tg.get_node_value(0, "task_group_id")
+    getitem_x_gid = tg.get_node_value(1, "task_group_id")
+    getitem_y_gid = tg.get_node_value(3, "task_group_id")
+    assert tup_electron_gid == 0
+    assert getitem_x_gid == tup_electron_gid
+    assert getitem_y_gid == tup_electron_gid
+    assert tg.get_node_value(2, "task_group_id") == 2
+    assert tg.get_node_value(4, "task_group_id") == 4
+    assert tg.get_node_value(5, "task_group_id") == 5
+    assert tg.get_node_value(6, "task_group_id") == 6
```
kessler-frost, line 498: This as well.

kessler-frost (changes requested):

covalent/_shared_files/defaults.py
FyzHsn: ah sorry they're not mutually exclusive as I realized recently - let me remove the comment

cjao: Agreed, `exhaustive_postprocess` alone should be sufficient.

covalent/_workflow/electron.py
cjao: That's a good point for `add_node`. The other difference is that only if `packing_tasks` is `False` should `self.task_group_id` be updated with the `node_id` of the newly added node; otherwise it should remain at the smallest `node_id` in the task group.

covalent/_workflow/postprocessing.py
```diff
@@ -173,3 +174,66 @@ def add_exhaustive_postprocess_node(self, bound_electrons: Dict) -> None:
...
+            # parameter node in case retval is not a single electron
+            bound_pp = pp_electron(retval, **referenced_electrons)
+
+            # Edit pp electron name
+            self.lattice.transport_graph.set_node_value(
+                bound_pp.node_id, "name", f"{postprocess_prefix}reconstruct"
+            )
+
+            # Wait for non-referenced electrons
+            if get_config("sdk.reconstruct_postprocess") != "true":
```
cjao, line 237: This isn't quite right -- the original condition was if an experimental `eager_postprocess` config option  was set, in which case postprocessing would kick off as soon as its dependencies were satisfied without waiting for other electrons. However that would cause a UX change as the workflow could be marked as "COMPLETED" before all electrons have finished executing (just the electrons in the connected component of the postprocessing electron).

Unless we want to expose such an "eager execution" option right now (@santoshkumarradha ), we should wait unconditionally for the non-referenced electrons so as to preserve the existing behavior in which postprocessing occurs as the last step in the workflow.

cjao (changes requested):

covalent/_workflow/postprocessing.py
cjao: I think "eager execution" would be mainly appropriate if users are constructing their workflow in a functional manner -- such that they only care about the workflow's final output. However if users are computing primarily with side effects, it would be wrong to declare a workflow as "COMPLETED" just because the workflow's output can be resolved.

cjao: "Eager execution" would benefit users who construct their workflow in a functional manner -- such that they only care about the workflow's final output. However if users are computing primarily with side effects, it would be wrong to declare a workflow as "COMPLETED" just because the workflow's output can be resolved.

covalent/_workflow/lattice.py
FyzHsn: Modified

kessler-frost (approved):

Issue #1584, opened 2023-04-20
Throw an error or warning if too many args/kwargs are provided at dispatch time


arosen93:
### What should we add?

Consider the example below:

```python3
import covalent as ct

@ct.lattice
@ct.electron
def job(val1, val2):
    return val1 + val2

dispatch_id = ct.dispatch(job)(1, 2, 10)
result = ct.get_result(dispatch_id)
print(result)
```

This runs successfully and returns the correct output. However, there are 3 positional args provided even though `Job` only takes 2. In complex workflows, this can lead to confusion for the end-user and potential bugs that might be hard to spot. My suggestion is that covalent should throw a warning or an error if extraneous args/kwargs are provided at dispatch time.

PR #1593, opened 2023-04-24
raise error on extra args/kwargs

madhur-tandon:
Fixes https://github.com/AgnostiqHQ/covalent/issues/1584


madhur-tandon:
> It might be worthwhile to try a more robust check with `inspect.signature` and let user know specifically which arguments or kwargs are missing or extra in the error as well.

So, I already tried `getfullargspec` stuff. There are two issues:

**1. It doesn't preserve the original function signature, probably because decorators change the metadata...**

Eg, if you try:

```
import covalent as ct

@ct.lattice
@ct.electron
def job(val1, val2):
    return val1 + val2
```

and do a `getfullargspec` on `func` inside the `get_named_params` function, we get:

```
FullArgSpec(args=[], varargs='args', varkw='kwargs', defaults=None, kwonlyargs=[],
kwonlydefaults=None, annotations={})
```

Clearly, the information about `val1` and `val2` is lost... while `func` still refers to `<function job at 0x106ae8b80>`

**2. It interferes with postprocessing...**

Specifically, `build_graph` calls `pp.add_exhaustive_postprocess_node` at some point which further uses the `_postprocess` function which has the signature `def _postprocess(self, *ordered_node_outputs: List[Any]) -> Any:`

Doing a `getfullargspec` on this gives us:

```
FullArgSpec(args=['self'], varargs='ordered_node_outputs', varkw=None, defaults=None, kwonlyargs=[],
kwonlydefaults=None, annotations={'return': typing.Any, 'ordered_node_outputs': typing.List[typing.Any]})
```

As one can see, we cannot use this information to count our `varargs` in some sense. Plus, it also includes `self` as one of the arguments (and will probably include `cls` too for class methods, if any).

Thus, even without using `getfullargspec` and directly counting with
`ordered_params_dict = inspect.signature(func).parameters` also leads to issues i.e. a simple comparison such as
`if len(args) + len(kwargs) > len(ordered_params_dict)` aka my first commit here: https://github.com/AgnostiqHQ/covalent/pull/1593/commits/79610c283c6b7e3fa81dbcbf6cace234d9e27752 will lead to errors...

---

In fact, the current approach of iterating and figuring out the `named_args` and `named_kwargs` is much better and hence, I resorted to using that itself...

My suggestion would be to tackle this "extended" functionality as a separate issue / PR owing to the above problems + the original issue also had a limited scope anyway...

kessler-frost:
Hey @madhur-tandon thanks for the PR 😄 ! I agree with the points you mentioned and that the current method of determining named args and kwargs is better, but I think what @santoshkumarradha mentioned:

> let user know specifically which arguments or kwargs are missing or extra in the error as well.

is quite achievable by only a few minor changes to the modifications you already made in the PR. For example, in the case of positional arguments we can tell the user how many extra args they have passed vs how many were expected. And for the keyword arguments we can tell them what were the keys that they passed which were not expected in the raised `ValueError`. Basically instead of comparing the length of all parameters, we split that into 2 `if` conditions one for `args` and one for `kwargs`.

Again, I realize the scope of the original issue did not include improvement in the error description, but the changes required to add that description shouldn't be a lot.

PS:
- You can use `set` for the key comparison: `set(kwargs.keys()) - set(named_kwargs.keys())`
- Bonus points if you can also handle the case where fewer arguments are passed than required. Something like this should help:
```python
        if param.default is param.empty and param_name not in kwargs and ind >= len(args):
            raise ValueError(f"Parameter {param_name} is required, but not supplied.")
```

covalent/_shared_files/utils.py
```diff
@@ -200,6 +202,10 @@ def get_named_params(func, args, kwargs):
                 if key != param_name:
                     named_kwargs[key] = value
 
+    actual_params_count = len(named_args) + len(named_kwargs)
+    if supplied_params_count > actual_params_count:
+        raise ValueError(f"Too many parameters given, expected {actual_params_count}")
+
```
kessler-frost, line 208:
```suggestion
    # If args are different than expected
    if len(args) > len(named_args):
        raise ValueError(f"Too many positional arguments given, expected {len(named_args)} but got {len(args)}")
    
    # If kwargs are different than expected
    if len(kwargs) > len(named_kwargs):
        diff = set(kwargs.keys()) - set(named_kwargs.keys())
        raise ValueError(f"Unexpected keyword arguments given: {diff}")

```

kessler-frost: Let me know what you think about the suggestion @madhur-tandon 😄 !

madhur-tandon:
@kessler-frost Sure, I TOO think that some improvements can be made with the current mechanism of figuring out the `named_args` and `named_kwargs`, I wrote the above text to:

- specifically highlight the issues with `getfullargspec` and that I had already tried it
- suggest that we could do the `extended` functionality in another PR

But, now that I think of it again, probably adding the same in this PR makes sense as well.

So, thanks for your comments, I will probably make some changes, add tests, and push again in some time...

kessler-frost:
Sounds good 🤘

madhur-tandon:
I suggest not to figure out the case where fewer arguments are present and the associated check of:

```
if param.default is param.empty and param_name not in kwargs and ind >= len(args):
    raise ValueError(f"Parameter {param_name} is required, but not supplied.")
```

Instead, since we call the function anyway at some point i.e. inside `build_graph`:

```
try:
    retval = workflow_function(*new_args, **new_kwargs)
```

The above call is enough to determine missing arguments and will throw an error on it's own i.e. it won't be thrown by us, but the purpose is served anyway.

I say because, considering the below snippet for example:

```
@ct.electron
def task(x, y, *z):
    return x + y + sum(z)

@ct.lattice
def workflow(x, y, *z):
    return task(x, y, *z)

workflow.build_graph(x=7, y=8)
```

There are no issues with this but then, the snippet for calculating fewer arguments would check for `z` here as well.

Thus, it is better to outsource this to the call of the function instead.

kessler-frost:
Sure, we can figure out the resolution for that in a separate issue then.

madhur-tandon:
@kessler-frost seems like there's an issue with CI and resolving the environment and dependencies, can you have a look and re-trigger the CI?

covalent/_shared_files/utils.py
madhur-tandon: accommodated, resolving...

tests/covalent_tests/workflow/lattice_test.py
```diff
@@ -107,3 +109,38 @@ def workflow(x):
...
+
+    with pytest.raises(ValueError, match="Unexpected keyword arguments: a, b"):
+        workflow.build_graph(1, a=2, b=3)
+
+    with pytest.raises(ValueError, match="Unexpected keyword arguments: a"):
+        workflow.build_graph(a=1)
+
+    # fewer arguments handled internally by function call
+    with pytest.raises(TypeError, match="missing 1 required positional argument: 'y'"):
+        workflow.build_graph(1)
```
kessler-frost, line 146: Quite comprehensive, I like it 💯 !

kessler-frost (approved):
Great job 😄!

Issue #1127, opened 2022-08-25
Re-enable `test_run_workflow_does_not_deserialize` test


kessler-frost:
### Environment

- **Covalent version**: 0.185.0
- **Python version**: 3.8.13
- **Operating system**: MacOS 12.3.1

### What is happening?

The `test_run_workflow_does_not_deserialize` test in `tests/covalent_dispatcher_tests/_core/execution_test.py` currently seems to be disabled due to a missing `pytest.mark.asyncio`. But even after enabling it or running it in a separate event loop, it doesn't seem to pass. In the former case the error says that the event loop is closed, and in the latter we get 

`TypeError: cannot pickle 'weakref' object`

### How can we reproduce the issue?

Try running the test with either `pytest.mark.asyncio` or in a separate event loop using `asyncio.run`.

### What should happen?

Mentioned test should pass.

PR #1612, opened 2023-05-01
Re-enable `test_run_workflow_does_not_deserialize`

madhur-tandon:
Fixes https://github.com/AgnostiqHQ/covalent/issues/1127


madhur-tandon:
The test passes if covalent dispatcher server is running, but the CI didn't start it, thus the test failure.

madhur-tandon:
Perhaps a few more mocks could fix this, looking...

madhur-tandon:
Could appreciate a review,
CC: @kessler-frost @santoshkumarradha @AlejandroEsquivel

tests/covalent_dispatcher_tests/_core/execution_test.py
```diff
@@ -37,6 +39,28 @@
...
+    )
+    yield cluster
+    cluster.close()
+
+
+@pytest.fixture
+def event_loop():
+    loop = asyncio.new_event_loop()
+    yield loop
+    loop.close()
```
kessler-frost, line 62: Are you sure we need this fixture? The event loop should get automatically passed to the asyncio marked test as mentioned here: https://pytest-asyncio.readthedocs.io/en/latest/concepts.html.

tests/covalent_dispatcher_tests/_core/execution_test.py
```diff
@@ -317,27 +339,42 @@ def workflow(x):
         res1 = ct.electron(sublattice_task(x), executor=dask_exec)
         return res1
 
+    asyncio.set_event_loop(event_loop)
+
```
kessler-frost, line 343: Here as well, I think the event loop will already be set by pytest unless I'm missing something here.

tests/covalent_dispatcher_tests/_core/execution_test.py
```diff
@@ -37,6 +39,28 @@
...
+@pytest.fixture
+def test_cluster():
+    cluster = LocalCluster(
+        n_workers=DEFAULT_N_WORKERS,
+        threads_per_worker=DEFAULT_THREADS_PER_WORKERS,
+        **{"memory_limit": DEFAULT_MEM_PER_WORKER},
+    )
+    yield cluster
+    cluster.close()
```
kessler-frost, line 55: Why don't we just return the cluster here instead of yielding? We can close it later. In fact, I think we can get away with just using a local executor here.

tests/covalent_dispatcher_tests/_core/execution_test.py
```diff
@@ -294,17 +318,15 @@ def workflow(x):
...
     out-of-process `workflow_executor`"""
 
-    from dask.distributed import LocalCluster
-
     from covalent._workflow.lattice import Lattice
     from covalent.executor import DaskExecutor
 
-    lc = LocalCluster()
-    dask_exec = DaskExecutor(lc.scheduler_address)
+    dask_exec = DaskExecutor(test_cluster.scheduler_address)
```
kessler-frost, line 329: Let's switch to a local executor, less stuff to handle.

tests/covalent_dispatcher_tests/_core/execution_test.py
```diff
@@ -294,17 +318,15 @@ def workflow(x):
...
-async def test_run_workflow_does_not_deserialize(mocker):
+@pytest.mark.asyncio
+async def test_run_workflow_does_not_deserialize(mocker, test_cluster, event_loop):
     """Check that dispatcher does not deserialize user data when using
     out-of-process `workflow_executor`"""
 
-    from dask.distributed import LocalCluster
-
     from covalent._workflow.lattice import Lattice
     from covalent.executor import DaskExecutor
```
kessler-frost, line 327: Let's also remove these from here, we're already importing `Lattice` at the top of the file.

kessler-frost (changes requested):
Thanks a lot @madhur-tandon ! Some small changes suggested 👍 .

tests/covalent_dispatcher_tests/_core/execution_test.py
cjao: @kessler-frost I remember this test being particularly brittle because pytest-asyncio likes to close the event loop prematurely.

tests/covalent_dispatcher_tests/_core/execution_test.py
cjao: Agreed. The original use for Dask was to run the tasks out-of-process, but LocalExecutor now does  that too.

madhur-tandon: Thanks, I will switch to using a local executor...

tests/covalent_dispatcher_tests/_core/execution_test.py
madhur-tandon: Done...

tests/covalent_dispatcher_tests/_core/execution_test.py
madhur-tandon: This was probably because of using a `DaskExecutor`.
I have switched to `LocalExecutor` though.

madhur-tandon:
@kessler-frost @cjao could appreciate a re-review...

tests/covalent_dispatcher_tests/_core/execution_test.py
```diff
@@ -294,50 +294,58 @@ def workflow(x):
...
+    @ct.lattice(executor="local", workflow_executor="local")
     def workflow(x):
         # Exercise both sublatticing and postprocessing
-        sublattice_task = ct.lattice(task, workflow_executor=dask_exec)
-        res1 = ct.electron(sublattice_task(x), executor=dask_exec)
+        sublattice_task = ct.lattice(task, workflow_executor="local")
+        res1 = ct.electron(sublattice_task(x), executor="local")
         return res1
 
+    # asyncio.set_event_loop(event_loop)
```
kessler-frost, line 313: Can you remove this line as well 😄 .

kessler-frost (changes requested):
Looks great 💯 ! Just one minor thing and then its gtg 🚀 !

tests/covalent_dispatcher_tests/_core/execution_test.py
madhur-tandon: Thanks, I must have forgotten about this 😅

kessler-frost (approved):

PR #1622, opened 2023-05-05
Fixed db path creation

### LINKED ISSUE #1621: Running any CLI command will fail on a new machine
### Environment

- **Covalent version**: 0.222.0rc0
- **Python version**: 3.8.15
- **Operating system**: Ubuntu 20.04

### What is happening?

Executing any CLI command throws an error now since it is unable to find the non-existent directory for the database, which it is looking for when because we are executing code during import time. This is why the tests are currently failing on `develop` branch.

After doing a `covalent purge -Hy` and deleting the `~/.local/share/covalent` directory if we run any command like `covalent db migrate` command, which is usually done as part of the test pipeline it will throw a long traceback with the final error being:

```python
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) unable to open database file
```

This happens because the database directory now only gets created after `covalent start` command is run.

### What should happen?

No error should be thrown when running a CLI command.

### Any suggestions?

Ideally, we should not be executing anything during import time (which happens to be the case for this issue), but resolving that might open up a different can of worms. So an alternative solution is to move the following line back to `ConfigManager` class' definition from its current place in `_graceful_start()`:

```python
Path(get_config("dispatcher.db_path")).parent.mkdir(parents=True, exist_ok=True)
```

kessler-frost:

Resolves #1621

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- DB path creation now takes place at import time so that the CLI commands don't fail
 - Raise error on dispatching a non-lattice
 - Helpful message when trying to dispatch a workflow when covalent server is not available
 - Open UI preview of transport graph when `lattice.draw()` is invoked and print URL of the same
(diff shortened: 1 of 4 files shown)


FyzHsn (approved):

madhur-tandon:
Thanks! This explains the CI failures.

PR #1631, opened 2023-05-07
mock set_config so as to not set port as 42 in CI

madhur-tandon:

This should hopefully help fix functional tests on CI...


madhur-tandon:
@kessler-frost 

Functional tests were failing in CI, only because there were being run after `covalent_dispatcher_tests` which includes the `service_test.py` file which uses port as `42`

Hence, functional tests start firing requests at the server at port `42`

When run in isolation though, all functional tests PASS, it is only in CI they fail because the previous `covalent_dispatcher_tests` changes the port…

kessler-frost:
Makes sense, thanks 👍 !

kessler-frost (approved):

PR #1632, opened 2023-05-08
Add Azure Batch and Google Batch Executors to RTD Index

arosen93:
Currently, the Azure Batch and Google Batch executor instructions aren't listed in the [Plugins index](https://covalent.readthedocs.io/en/latest/plugins.html), meaning that most users probably don't know about it. This should fix it.

There are a few additional fixes to be done in a separate PR, as noted in #1633.


arosen93:
Feel free to merge whenever you like.

arosen93:
Closed because this was a bit premature to make public. See #1633.

PR #1640, opened 2023-05-11
Fix docstring for `set_config`

### LINKED ISSUE #1639: Incorrect docstring for `update_config`
### What should we add?

https://github.com/AgnostiqHQ/covalent/blob/01738d33b2eead05d59cea5fa5d0365c4ac1bead/covalent/_shared_files/config.py#L274-L286

The default for `override_existing` is `True`, but the docstring says `False` (for a kwarg `defaults` that doesn' exist).

arosen93:
Closes #1639.

- [NA] I have added the tests to cover my changes.


santoshkumarradha (approved):
@arosen93 🙏 thanks again for the catch!

PR #1660, opened 2023-05-22
Adding `executor` as a property

### LINKED ISSUE #1647: Ability to update the executor of a pre-defined or imported `Electron` object
### What should we add?

There should be some clean way for the user to update the executor of an `Electron` once it is has been defined or imported from some external Python package. See #1646 for the full context.

@santoshkumarradha has good suggestions here: https://github.com/AgnostiqHQ/covalent/discussions/1646#discussioncomment-5901761

> hey @arosen93 , this is a long backlogged item on our plate. Ideally the best UX is to have something like
> 
> ### UX1
> ```python
> from my_package import energy_electron # imports a direct electron object
> 
> energy_electron.executor=MyExecutor()
> ```
> 
> Could @kessler-frost confirm if using `energy_electron.metadata["executor"]=MyExecutor()` would be a valid solution? I'm uncertain if it would properly set everything needed. I recall encountering some issues with UX1 in the past, but perhaps they have since been resolved and we can revisit the matter.
> 
> Meanwhile I am just curious if using a UX2 would be a good fit for creating custom electron packages.
> 
> ### UX2
> ```python
> from my_package import energy_electron
> small_cell=energy_electron(**parameters,executor=MyExecutor(num_cpu=2)) # Returns an electron function 
> super_cell=energy_electron(**parameters,executor=MyExecutor(num_cpu=45)) # Returns an electron function 
> 
> @ct.lattice
> def workflow():
>     energy_supercell=super_cell("NaCl") #different resource electron
>     energy_smallcell=small_cell("NaCl") #different resource electron
> ```
> 
> In certain scenarios, UX2 may be more suitable as it allows for the creation of multiple electron instances with unique parameters through a factory pattern. This is not achievable in UX1, which is designed to return a singleton electron instance with a specific tied instance. Infact our initial interest in UX1 was coming in from a perspective of changing executor inside an interactive notebook after electrons have been defined.
> 
> LMK your thoughts.

Modifying the `electron_object.metadata` directly, but this doesn't seem like a "suggested" approach.

kessler-frost:

Resolves #1647

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+### Added
+
+- `executor` property to `Electron` class, allowing updation of executor after electron function definition
+
 ### Operations
 
 - Nightly frequency set to midnight EST 
--- a/covalent/_workflow/electron.py
+++ b/covalent/_workflow/electron.py
@@ -386,9 +386,7 @@ class Electron:
                 and k in DEFAULT_METADATA_VALUES
                 and not self.get_metadata(k)
             ):
-                meta = active_lattice.get_metadata(k)
-                if not meta:
-                    meta = DEFAULT_METADATA_VALUES[k]
+                meta = active_lattice.get_metadata(k) or DEFAULT_METADATA_VALUES[k]
                 self.set_metadata(k, meta)
 
         # Handle sublattices by injecting _build_sublattice_graph node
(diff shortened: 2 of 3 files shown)


kessler-frost:
Something like this should work now:

```python
import covalent as ct

# Custom package containing custom electron
# my_package.py:
@ct.electron
def my_custom_electron(a, b):
    return "May the force be with you." if a == b else "Hello World!"

custom_task = my_custom_electron.electron_object

# In a separate file or notebook:
# from my_package import custom_task

custom_task.executor = "local"

@ct.lattice
def my_workflow():
    return custom_task(4, 4)

print(ct.dispatch_sync(my_workflow)())
```
I've only added support for executor for now, but adding support for other constraints like `deps`, `call_before`, etc. is equally simple and can be done if need arises. Lmk your thoughts @santoshkumarradha @arosen93 .

arosen93:
Very clean! Thanks for adding this!

santoshkumarradha (approved):
Works perfectly!

kessler-frost:
Ah, you're right @cjao. I had a feeling I was missing something. Thanks 🤘. I'll create a follow up PR for this.

santoshkumarradha:
@kessler-frost I see that the fix is quite trivial to just add it directly inside the decorator, could you pass a PR for this one ?

PR #1601, opened 2023-04-25
dispatch when covalent server is not available

madhur-tandon:
Fixes https://github.com/AgnostiqHQ/covalent/issues/1260

- [x] I have added the tests to cover my changes.
- [x] I have updated the documentation and CHANGELOG accordingly.
- [x] I have read the CONTRIBUTING document.

PR #1662, opened 2023-05-23
Enhancing error messages

wjcunningham7:
This PR addresses a comment by @arosen93 about how to better handle the scenario where a server is not started or otherwise unreachable.  Now, the methods `dispatch`, `redispatch`, and `get_result` will print a helpful error message and return `None` if this is encountered.

To test these changes, invoke one of these methods while the server is stopped.


madhur-tandon:
@wjcunningham7 you might have to adjust some tests...
I remember adding `test_dispatch_when_no_server_is_running` in https://github.com/AgnostiqHQ/covalent/pull/1601
But there might be others...

Issue #1670, opened 2023-05-26
Make `setup.py` OS-agnostic and replace `fcntl` with an OS-agnostic alternative


arosen93:
### What should we add?

Setting aside some potential UI challenges, if you want to make Covalent `Electron` objects compatible with Windows (e.g. to use as regular functions), I don't think it's all that difficult.

---------

The first change is to modify `open("README.md").read()` to be `open("README.md", encoding="utf8").read()` below so the Windows user can `pip install covalent`.

https://github.com/AgnostiqHQ/covalent/blob/90d1560a5d9324aa7d67f28ff1864f4de65b06b5/setup.py#L209

**Edit**: This was merged in #1685.

---------

The second change is to use an OS-agnostic locking mechanism in place of `fcntl` below so the config file can be properly written. [Portalocker](https://github.com/wolph/portalocker) is one such option, which is basically a drop-in replacement of `fcntl.lockf(f, fcntl.LOCK_EX)` with `portalocker.lock(f, portalocker.LOCK_EX)`.

https://github.com/AgnostiqHQ/covalent/blob/90d1560a5d9324aa7d67f28ff1864f4de65b06b5/covalent/_shared_files/config.py#L112-L148

-------

Once those two changes are made, you can import/use `Electron` objects as normal functions on Windows.

sgbaird:
`covalent start` on Windows 11 for me results in:

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\sterg\miniconda3\envs\self-driving-lab-demo\Scripts\covalent.exe\__main__.py", line 4, in <module>
  File "C:\Users\sterg\miniconda3\envs\self-driving-lab-demo\Lib\site-packages\covalent_dispatcher\__init__.py", line 17, in <module>    
    from .entry_point import cancel_running_dispatch, run_dispatcher
  File "C:\Users\sterg\miniconda3\envs\self-driving-lab-demo\Lib\site-packages\covalent_dispatcher\entry_point.py", line 24, in <module> 
    from covalent._shared_files import logger
  File "C:\Users\sterg\miniconda3\envs\self-driving-lab-demo\Lib\site-packages\covalent\__init__.py", line 22, in <module>
    from . import executor, leptons  # nopycln: import
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1685, opened 2023-06-04
Correct the docstring for the default executor in `Lattice`

### LINKED ISSUE #1682: Incorrect docstring on `Lattice` object
### What should we add?

https://covalent.readthedocs.io/en/latest/api/lattice.html?highlight=metadata#covalent.lattice

> executor – Alternative executor object to be used in the execution of each node. If not passed, the local executor is used by default.

I think the default should be the Dask executor, if I understand correctly.

arosen93:
- Closes #1682: corrects the `Lattice` docstring to say the Dask executor is default and not Local
- Addresses part of #1670: reads in the `README.md` in `setup.py` as `encoding="utf-8"`


kessler-frost (approved):
Thanks @arosen93 ! Looks gtg 🚀 !

santoshkumarradha (dismissed):

AlejandroEsquivel (approved):

PR #1694, opened 2023-06-08
adding azure blob storage file transfer strategy

wjcunningham7:
This pull request adds file transfer support for Azure blob storage.  An example usage is shown below:

```
import covalent as ct 

blob_strategy = ct.fs_strategies.Blob(
    client_id=<sp_client_id>,
    client_secret=<sp_client_secret>,
    tenant_id=<my_tenant_id>,
)

input_file = ct.fs.TransferFromRemote(
    <remote_path>,
    <local_path>,
    strategy=blob_strategy,
)

output_file = ct.fs.TransferToRemote(
    <remote_path>,
    <local_path>,
    strategy=blob_strategy,
)

@ct.electron(files=[input_file, output_file])
def blob_transfer(files):
    import os

    file_path = files[0][1]
    return os.path.exists(file_path)

@ct.lattice
def test_blob_transfers():
    return blob_transfer()

ct.dispatch(test_blob_transfers)()
```

Additionally, if any file source specified in `TransferFromRemote` or `TransferToRemote` ends in a trailing slash (`/`) then it will be recognized as a directory download/upload and handled appropriately.  In order to use this feature, we assume the user has previously created a Blob storage account, a storage container within the storage account, service principal, and credentials for the service principal.  The service principal must have read-write access to the storage container.

EDIT: For Azure Blob URIs, we assume the following standard format:
```
https://<storage_account_name>.blob.core.windows.net/<container_name>/<blob_name>
```


covalent/_file_transfer/enums.py
```diff
@@ -29,6 +29,7 @@ class Order(str, enum.Enum):
 class FileSchemes(str, enum.Enum):
     File = "file"
     S3 = "s3"
+    Blob = "blob"
```
cjao, line 32: Is the Azure blob URI scheme "blob" or ["abfss"](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri)?

wjcunningham7: It looks like it uses [https](https://learn.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata) actually; we would need a separate interface for ADLS.  We can either use `blob` to distinguish these paths from generic `https` or stick with https and discourage scheme inference in the case of Azure blob storage.

wjcunningham7: @santoshkumarradha

wjcunningham7: current method filters by https and also checks for `blob.core.windows.net` in the `host` portion of the URI in order to determine a Blob transfer strategy.

cjao (approved):

PR #1699, opened 2023-06-12
Add google storage file transfer strategy

wjcunningham7:
This pull request adds file transfer support for Google Cloud storage.  To use this users will need to have set up a bucket in a Google Cloud storage account (with a billing account attached).  Additionally users will need to have created a service account + keys for machine-based authentication.  During testing I used the `Storage Admin` role for the service account.  

Syntax is the following:

```
import covalent as ct

gcloud_strategy = ct.fs_strategies.GCloud(
    credentials="/path/to/service_account_credentials.json",
    project_id="my_gcp_project_id",
)

input_file = ct.fs.TransferFromRemote(
    "gs://<remote-bucket>/<filename>",
    "/local/path/to/downloaded/object",
    strategy=gcloud_strategy,
)
output_file = ct.fs.TransferToRemote(
    "gs://<remote-bucket>/<uploaded_filename>",
    "/local/path/to/object",
    strategy=gcloud_strategy,
)

@ct.electron(files=[input_file, output_file])
def gcloud_transfer(files):
    file_path = files[0][1]
    return os.path.exists(file_path)

@ct.lattice
def test_gcloud_transfers():
    return gcloud_transfer()

dispatch_id = ct.dispatch(test_gcloud_transfers)()
print(dispatch_id)
```

Note that we assume URIs in Google Cloud Storage use the format `gs://<bucket_name>/<blob_name>` and recognize `gs` as the file scheme of remote objects.


covalent/_file_transfer/enums.py
```diff
@@ -49,6 +51,7 @@ class FileTransferStrategyTypes(str, enum.Enum):
     "http": FileTransferStrategyTypes.HTTP,
     "https": FileTransferStrategyTypes.HTTP,
     "s3": FileTransferStrategyTypes.S3,
+    "gs": FileTransferStrategyTypes.GCloud,
```
cjao, line 54: @wjcunningham7 I think I can now better articulate my earlier comment on the Azure PR -- since the scheme is normally identified by the URL prefix (as done here for example), it is nonstandard to have "https" refer to both HTTPS and and Blob schemes.

wjcunningham7: I agree with that but I don't know of a good solution that doesn't introduce something else nonstandard.

AlejandroEsquivel (approved):
Looks good to me @wjcunningham7. 

It seems that the upload and download files feature currently uploads/downloads files synchronously however I don't think any of the other file transfer strategies currently do parallel uploading/downloading so I don't think this would be an issue. Although if there are large files to be uploaded and downloaded, as these happen synchronously this might block the main thread and the covalent server may become unresponsive for that time. May be a later consideration.

PR #1654, opened 2023-05-17
only pin aiohttp downwards for py3.11 compat

janosh:
Running

```sh
pip install covalent
```

fails on Python 3.11:

```
      copying aiohttp/.hash/hdrs.py.hash -> build/lib.macosx-10.9-universal2-cpython-311/aiohttp/.hash
      running build_ext
      building 'aiohttp._websocket' extension
      creating build/temp.macosx-10.9-universal2-cpython-311
      creating build/temp.macosx-10.9-universal2-cpython-311/aiohttp
      clang -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -arch arm64 -arch x86_64 -g -I/Users/janosh/.venv/py311/include -I/Library/Frameworks/Python.framework/Versions/3.11/include/python3.11 -c aiohttp/_websocket.c -o build/temp.macosx-10.9-universal2-cpython-311/aiohttp/_websocket.o
      aiohttp/_websocket.c:198:12: fatal error: 'longintrepr.h' file not found
        #include "longintrepr.h"
                 ^~~~~~~~~~~~~~~
      1 error generated.
      error: command '/usr/bin/clang' failed with exit code 1
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for aiohttp
Failed to build aiohttp
ERROR: Could not build wheels for aiohttp, which is required to install pyproject.toml-based projects
```

Looks like [`aiohttp==3.8.2`](https://github.com/aio-libs/aiohttp/issues/6600) fixed install on 3.11 so removing the strict pin should fix this.

PR #1702, opened 2023-06-14
Reduce pins in `setup.py` per standard Python recommendations

### LINKED ISSUE #1684: Minimize number of pinned versions in the `install_requires` of `setup.py`
### What should we add?

Currently, the `setup.py` file pulls the packages from `requirements.txt` in the `install_requires`. To increase the compatibility of Covalent with other packages and to abide by standard Python recommendations, `setup.py` should only have pins in the `install_requires` where necessary. `requirements.txt` is where the pinned versions should exist to reproduce the testing environment.

[Source](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/): 

> It is not considered best practice to use install_requires to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades.

This was alluded to in #1654, but I've now opened a proper issue for it.

Alternatively, the `requirements.txt` could be removed entirely, the `install_requires` in `setup.py` could have versions mostly unpinned, and a new `[strict]` extras could be added with pinned versions for testing. This approach is less common, however.

arosen93:
Closes #1684 (and addresses the concern brought up in #1654).

> It is not considered best practice to use install_requires to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades. [Source](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/)

Pins to reproduce a given test environment are meant to be stored in `requirements.txt`, but having them in `setup.py` makes it more difficult on developers who may plan to use Covalent as a dependency. The test suites should install from `requirements.txt`, but the user should be able to have more flexible versions unless it's _known_ that a given dependency causes issues with certain versions. Dependabot can then be used to update any strict-equals pins in both `setup.py` (if any) and `requirements.txt`. Since the test suite is run on `requirements.txt`, any time a version update causes a conflict, an upper bound can be easily set in `setup.py` as needed.

To confirm that the proposed `setup.py` `install_requires` are indeed suitable as I've outlined them here, I have ran the test suite on the dependencies from `setup.py` rather than `requirements.txt` and ensured that all tests still pass.

See the test [here](https://github.com/AgnostiqHQ/covalent/actions/runs/5264978642/jobs/9516992306) from commit [35a7c72ffee0ce696a65dd0f8092721f8e5b2a5a](https://github.com/AgnostiqHQ/covalent/tree/35a7c72ffee0ce696a65dd0f8092721f8e5b2a5a) for evidence that the proposed `setup.py` deps keep the test suite happy. The only package with a strict upper limit right now is `uvicorn[standard]`.


arosen93:
I think this should be ready for review now! :)

kessler-frost:
Thanks for this @arosen93 ! I'll take look at this tomorrow (Monday)!

requirements.txt
```diff
@@ -1,20 +1,20 @@
...
 simplejson==3.17.6
-sqlalchemy>=1.4.37,<=1.4.41
+sqlalchemy==1.4.41
 sqlalchemy_utils==0.38.3
 toml==0.10.2
-uvicorn[standard]>=0.18.2,<=0.18.3
+uvicorn[standard]==0.18.3
 watchdog==2.2.1
-werkzeug>=2.0.3,<=2.2.2
+werkzeug==2.2.2
```
kessler-frost, line 20: Instead of exactly pinning the versions like this, let's use the [compatible release](https://peps.python.org/pep-0440/#compatible-release) operator so that we're not tied to one version and upgrades are ensured to work due to backwards compatibility.

If it's not too much could you also edit the `tests/requirements.txt` file for the same reason. Much appreciated, thanks 🙏 !

setup.py
```diff
@@ -210,7 +207,28 @@ def find_sources(self):
...
+        "python-socketio",
+        "requests>=2.24.0",
+        "simplejson",
+        "sqlalchemy>=1.4.37",
+        "sqlalchemy_utils",
+        "toml",
+        "uvicorn[standard]==0.18.3",
+        "watchdog",
+        "werkzeug>=2.0.3",
+    ],
```
kessler-frost, line 231: Let's put this entire list in a variable instead somewhere early in the file. Will make adding/updating things easier.

kessler-frost (changes requested):
Thanks @arosen93 ! Only a few minor things, otherwise looks gtg 🚀 !

requirements.txt
kessler-frost: In fact, I'm almost leaning towards not using the strict pins at all here and only use the `>=` one. Once/If something breaks due to the dependency upgrade, then we can put it within a range.

kessler-frost: @arosen93 that's a good point. Although, I'd say that ideally any breakage in dependency should get caught in the nightly run of the test suite that takes place at around 12 am EST 😅 . Plus, we'd also like to know the reason behind that breakage, and whether it is better to improve the code which is using that broken dependency or just restrict the version of that dependency - which in itself might warrant a manual PR.

arosen93: @kessler-frost, ah that makes a lot of sense! I forgot about the nightly test.

So, one final question then. If there's no plan to have version pins in the `requirements.txt`, is there a benefit of having a separate `install_requires` in the `setup.py` after all? Would the best solution to read the `install_requires` from `requirements.txt` like what's in the `develop` branch but to use the less strict versions I've include in the `setup.py` within this PR?

kessler-frost: Ah, yeah so your issue (and this PR) actually sparked a discussion which led us to consider using a `requirements-ci.txt` instead in the near future. Which, I'd say this `requirements.txt` will either transition into or be a supplement to this. In `install_requires` eventually what we'd like to do is have no version restrictions at all for as many packages as possible. So, until then we should have both `install_requires` and `requirements.txt` which can evolve independently 😄 .

kessler-frost (dismissed):
Thanks a lot @arosen93 🚀  !

wjcunningham7:
@arosen93 I'm happy to remove strict pins but I think we want to be careful here so that we don't introduce any instabilities.  I'd be ok with removing upper bounds but the lower bounds are important as they represent what's known to be working (`networkx` is a good example of this).

With regards to the changes in `setup.py`, I'd prefer to keep the current style:
- We are currently using `requirements.txt` as a place to store dependencies, not necessarily for a super-specific environment used in CI or testing.  Since it's read in when `setup.py` is imported, there's no functional difference here, just stylistic.  Any "extras" for CI are included in `tests/requirements.txt`.
- It's important that `requirements-client.txt` is respected when `COVALENT_SDK_ONLY=1` so that SDK containers and distributions remain small.  We could optionally include a second list in `setup.py` to cover this, but then we have 4 places dependencies need to be updated.
- It is important to include at least minimum versions to ensure a smooth upgrade process across stable versions of Covalent.

arosen93:
@wjcunningham7: That certainly makes sense! A few comments/questions in return though.

> I'd be ok with removing upper bounds but the lower bounds are important as they represent what's known to be working (networkx is a good example of this).

I think I only removed the upper bounds. I don't see any lower bounds that I removed. If I removed any lower bounds, that's a mistake that should be fixed. The `networkx` example was a strict pin before without a lower bound. Is your suggestion to add all strict pins as lower bounds in cases where there are no pins in the proposed `setup.py`? For instance, `furl>=2.1.3` and `networkx>=2.8.6`? That can certainly be done, although of course it's possible that earlier versions are also supported. I will make this change, but correct me if I'm mistaken.

> We are currently using requirements.txt as a place to store dependencies, not necessarily for a super-specific environment used in CI or testing. Since it's read in when setup.py is imported, there's no functional difference here, just stylistic. Any "extras" for CI are included in tests/requirements.txt.

Certainly, we can go with that route. I agree that having effectively duplicated `setup.py` and `requirements.txt` files is not ideal. It should either be that `requirements.txt` is strict pins and `setup.py` is not, or `setup.py` should be reading from `requirements.txt` --- not both. I can revert this change.

> It's important that requirements-client.txt is respected when COVALENT_SDK_ONLY=1 so that SDK containers and distributions remain small. We could optionally include a second list in setup.py to cover this, but then we have 4 places dependencies need to be updated.

Sorry, could you clarify what you would prefer to happen with respect to `requirements-client.txt`? Thanks!

-----

Overall, happy to do whatever you prefer! My main goal is just reduce the number of strictly pinned versions in `install_requires` when there is no reason for them to be pinned.

arosen93:
@wjcunningham7, is this closer to what you were looking for? Happy to make any other changes you see fit.

wjcunningham7:
> @wjcunningham7, is this closer to what you were looking for? Happy to make any other changes you see fit.

Yeah this is perfect.  I think this strikes the right balance between flexibility and stability.

wjcunningham7 (approved):

arosen93:
Beautiful. Thanks!

PR #1705, opened 2023-06-20
Remove duplicate "stop server" warning in docs

arosen93:
The warning "Do not stop the Covalent server while you have running workflows. Stopping the server will kill the workflows." was included twice on [this page](https://covalent.readthedocs.io/en/latest/getting_started/first_experiment/index.html). The second instance doesn't seem like the best spot to include it, because it might scare the user into thinking closing their Python console or killing their Jupyter kernel will kill the server.

- [X] I have added the tests to cover my changes.


santoshkumarradha (approved):
I agree thanks for improving the docs as always @arosen93 !!! 

🚀

PR #1719, opened 2023-07-04
Fix for config file locking

kessler-frost:


arosen93:
Closes #1670.

arosen93:
By the way, not that it needs to be done in this PR, but there's one other section in the codebase that relies on the OS-dependent `fcntl`.

https://github.com/AgnostiqHQ/covalent/blob/691f0ebf57ca6fcfbe48a3b6cd5e2bb79b178400/covalent_ui/api/main.py#L73-L75

Maybe it'll be possible to replace that call too at some point.

kessler-frost:
Yep that's on my radar. Thanks @arosen93 !

kessler-frost:
@arosen93 Do you mind helping me out a bit with this one, can you check whether this solution works on your side on the slurm cluster for which covalent was halting earlier? Thanks a lot!

arosen93:
@kessler-frost: I reran my typical test. A few notes.

1. After pip install-ing, it didn't install filelock. So, I had to do that manually.
2. Covalent no longer freezes on import
3. Jobs do not run successfully by default (traceback below), _but this is expected._

```python
  File "/pscratch/sd/r/rosen/test.py", line 1, in <module>
    import covalent
  File "/global/common/software/matgen/rosen/miniconda/envs/test/lib/python3.10/site-packages/covalent/__init__.py", line 26, in <module>
    from . import executor, leptons  # nopycln: import
  File "/global/common/software/matgen/rosen/miniconda/envs/test/lib/python3.10/site-packages/covalent/executor/__init__.py", line 33, in <module>
    from .._shared_files import logger
  File "/global/common/software/matgen/rosen/miniconda/envs/test/lib/python3.10/site-packages/covalent/_shared_files/logger.py", line 34, in <module>
    log_level = get_config("sdk.log_level").upper()
  File "/global/common/software/matgen/rosen/miniconda/envs/test/lib/python3.10/site-packages/covalent/_shared_files/config.py", line 244, in get_config
    cm = ConfigManager()
  File "/global/common/software/matgen/rosen/miniconda/envs/test/lib/python3.10/site-packages/covalent/_shared_files/config.py", line 56, in __init__
    self.update_config()
  File "/global/common/software/matgen/rosen/miniconda/envs/test/lib/python3.10/site-packages/covalent/_shared_files/config.py", line 113, in update_config
    with filelock.FileLock(f"{self.config_file}.lock", timeout=1):
  File "/global/common/software/matgen/rosen/miniconda/envs/test/lib/python3.10/site-packages/filelock/_api.py", line 255, in __enter__
    self.acquire()
  File "/global/common/software/matgen/rosen/miniconda/envs/test/lib/python3.10/site-packages/filelock/_api.py", line 222, in acquire
    raise Timeout(lock_filename)  # noqa: TRY301
filelock._error.Timeout: The file lock '/global/homes/r/rosen/.config/covalent/covalent.conf.lock' could not be acquired.
```

The reason the lock can't be acquired is because the default `COVALENT_CONFIG_DIR` (being in `~/`) on Perlmutter doesn't supporting locking mechanisms in general. I don't see a real way of getting around that aside from not using a locking mechanism at all. But more importantly, because there's no stalling and a _proper_ error is raised, at least it is telling the user to do something about it. In this case, the user has to just update their `COVALENT_CONFIG_DIR` to a filesystem that supports locking. I've confirmed that everything works as expected once that's done.

So, in summary, I think this solution is better and worth merging because:
- There is no stalling upon error
- A proper error is raised
- It is an OS-agnostic solution

Just make sure `filelock` makes its way into `setup.py`.

kessler-frost:
Thanks a lot @arosen93 ! I'll clean this one up and get this merged.

arosen93:
Great! Thanks for addressing this :)

kessler-frost:
```python
 def set_winsize(fd, row, col, xpix=0, ypix=0): 
     winsize = struct.pack("HHHH", row, col, xpix, ypix) 
     fcntl.ioctl(fd, termios.TIOCSWINSZ, winsize) 
```
The above `fcntl` in `covalent/covalent_ui/api/main.py` can be changed to something else later since it's UI based and requires a different type of testing.

PR #1729, opened 2023-07-12
Memory improvements (2/3): Migrate core Covalent to new data access layer

cjao:

Closes https://github.com/AgnostiqHQ/covalent-staging/issues/708


cjao:
This PR migrates the core Covalent server from the in-memory Result object to the DB-backed `Result` object (based on `DispatchedObject`) introduced in the preceding PR. Dispatches are imported and exported through the new data access layer. The dispatcher is also made aware of task groups; however, as task-packing is not yet compatible with executor plugins, it remains disabled by default in the SDK settings.

kessler-frost (approved):

wjcunningham7 (approved):

Issue #1733, opened 2023-07-24
Setting the `executor` kwarg in `@ct.lattice` does not pull the right configuration parameters


arosen93:
### Environment

- **Covalent version**:  0.226.0-rc.0
- **Python version**: 3.10.12
- **Operating system**: Ubuntu

### What is happening?

This issue is largely a mirror of https://github.com/AgnostiqHQ/covalent-slurm-plugin/issues/74, but I feel like the cause may be independent of the plugin so am posting here too.

Basically, when passing in the `executor` keyword argument to the `@ct.lattice` decorator, there are situations where the configuration parameters of the `executor` object are being ignored in place of those found in the covalent configuration file.

### How can we reproduce the issue?

Start from a clean Covalent configuration. Then run the following, with the `slurm-covalent-plugin` installed.

```python
import covalent as ct

executor = ct.executor.SlurmExecutor(
    username="user", address="login.cluster.org", ssh_key_file="~/.ssh/id_rsa"
)

@ct.electron  # (executor=executor) works fine!
def add(a, b):
    return a + b
    
@ct.lattice(executor=executor)
def workflow():
    return add(1, 2)

dispatch_id = ct.dispatch(workflow)()
```

### What should happen?

The username should be registered as "user". However, the error message received is "username is a required parameter in the Slurm plugin" because it is being interpreted as the default value and because I am starting from a clean Covalent configuration where `username` has not been set in the configuration file.

This does not happen when the `executor` kwarg is passed to the `@ct.electron` decorator --- only the `@ct.lattice` decorator.

arosen93:
Closed in #1736! Yay!

PR #1736, opened 2023-07-25
fixing executor data inheritance

wjcunningham7:
This PR fixes a bug related to executor attribute inheritance.  Previously, if an executor class object was assigned to a lattice's executor or workflow_executor, only the string name would be passed to sublattices, leading to errors such as #1733 and https://github.com/AgnostiqHQ/covalent-slurm-plugin/issues/70.  This small change fixes this so that the dictionary of class metadata is also passed to sublattices.

After this fix, the following examples should work:

1. Executor inheritance
```
# MyExecutor is used for all electrons at any level of nesting
# Dask will be used as the workflow executor

executor = MyExecutor(param1, param2)

@ct.electron
def task():
    return

@ct.electron
@ct.lattice
def sublattice():
    return task()

@ct.lattice(executor=executor)
def workflow():
    return sublattice()
```

2. Workflow executor inheritance
```
# Both the top-level lattice and the sublattice use MyExecutor for postprocessing
@ct.lattice(workflow_executor=executor)
def workflow():
    return sublattice()
```

3. Override the default executor in a sublattice
```
# The sublattice's tasks will now use the local executor
# Any further levels of nesting will also use "local"
@ct.electron
@ct.lattice(executor="local")
def sublattice():
    return task()

@ct.lattice(executor=executor)
def workflow():
    return sublattice()
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reverted file-lock changes
 - Fixed dispatches list UI api caused by pydantic config.
 - Fixed graph API.
+- Executor and workflow executor data dictionaries are passed to sublattices
 
 ## [0.227.0-rc.0] - 2023-06-13
 
--- a/covalent/_shared_files/defaults.py
+++ b/covalent/_shared_files/defaults.py
@@ -178,7 +178,9 @@ class DefaultConfig:
 @dataclass
 class DefaultMetadataValues:
     executor: str = field(default_factory=get_default_executor)
+    executor_data: Dict = field(default_factory=dict)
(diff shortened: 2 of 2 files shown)


cjao (approved):

santoshkumarradha:
@arosen93 meanwhile quick note here building the UI is pretty straight [forward as well ](https://github.com/AgnostiqHQ/covalent/blob/develop/CONTRIBUTING.md#build-web-app) :D  I suspect this would close https://github.com/AgnostiqHQ/covalent-slurm-plugin/issues/69 as well, but we will do a quick smoke test.

arosen93:
@santoshkumarradha: Thanks! Yup, I was able to build the UI to close #1733! Haven't had a moment to test the sublattice issue on Slurm just yet but I imagine it'll be fixed too. 🎉

PR #1749, opened 2023-08-16
Quantum Tasks with Quantum Electrons Feature

AlejandroEsquivel:
This PR introduces the Quantum Tasks feature, integrating Quantum Electrons (QElectrons) into Classical Electrons via https://github.com/PennyLaneAI/pennylane/ Qnodes. Key requirements include:

- [x] **Distributed Execution**: QElectrons support distributed execution of circuit calculations.
- [x] **Resource Management**: Provides control over quantum resources within standard workflow Electrons.
- [x] **Observability and Monitoring**: Integrated system for caching, monitoring, and logging quantum tasks.
- [x] Hybrid Computation Support: Integration of QElectrons inside Electrons for executing quantum and classical sub-tasks on various backends.
- [x] **Quantum Resource Clustering**: Capability to form Quantum Clusters across different resource providers.
Custom Quantum Scheduling: Allows dynamic circuit allocation based on user-defined criteria.
- [x] A**synchronous Execution**: Support for non-blocking calls to Pennylane quantum circuits.
- [x] **Data Caching**: Quantum resource data caching mechanism implemented.
- [x] **UI Integration**: Quantum task monitoring and data inspection integrated into the Covalent UI.
Token Management Interface: Centralized interface for managing vendor-specific tokens.

------


santoshkumarradha (dismissed):

covalent/_results_manager/result.py
```diff
@@ -372,6 +375,27 @@ def _get_node_output(self, node_id: int) -> Any:
...
+            The QElectron data of said node. Will return None if no data exists.
+        """
+        try:
+            # Checks existence of QElectron data.
+            self._lattice.transport_graph.get_node_value(node_id, "qelectron_data_exists")
+        except KeyError:
+            return None
+
+        results_dir = get_config("dispatcher")["results_dir"]
+        db_dir = os.path.join(results_dir, self.dispatch_id, QE_DB_DIRNAME)
```
cjao, line 395: @kessler-frost does this look directly into the server's filesystem? Will it work with a self-hosted deployment?

covalent_dispatcher/_core/runner.py
```diff
@@ -243,6 +250,24 @@ async def _run_task(
...
         app_log.debug(f"Executing task {node_name}")
+
+        def qelectron_compatible_wrapper(node_id, dispatch_id, ser_user_fn, *args, **kwargs):
+
+            user_fn = ser_user_fn.get_deserialized()
+
+            try:
+                mod_qe_utils = importlib.import_module("covalent._shared_files.qelectron_utils")
+
+                with set_context(node_id, dispatch_id):
```
cjao, line 261: Since nearly all of the AWS plugins will need an update to actually support qelectrons, could we consider less invasive approaches to setting the context and retrieving the qelectron db? For instance, execution context could be injected into environment variables.

cjao: For the qelectron db, we could introduce another StringIO or BytesIO attribute in the BaseExecutor and ask the plugin to write the db to that buffer in its `run()` method similarly to how it is expected to return stdout and stderr. The runner-executor interface could then be extended in a backward compatible way to receive the extra task artifact.

cjao: This way, `generate_node_result()` remains lightweight as it simply assembles a dictionary instead of postprocessing a large `stdout`.

covalent/_results_manager/result.py
kessler-frost: It does access the server's filesystem directly. Which means this won't work in self-hosted deployment or in the case of cloud executors.

kessler-frost:
pre-commit.ci autofix

kessler-frost:
pre-commit.ci autofix

covalent/quantum/qserver/utils.py
```diff
@@ -0,0 +1,81 @@
...
+from pydantic import BaseModel
+
+from ..._shared_files.qinfo import QNodeSpecs
+from ...executor.qbase import BaseQExecutor
+
+BATCH_ID_SEPARATOR = "@"
+MAX_DIFFERENT_EXECUTORS = 10
+
+
+class CircuitInfo(BaseModel):
```
araghukas, line 36: Hey @kessler-frost, we also need to have `Optional[]` here, no?

Seems like these fields usually get populated with valid values, but if you call a QElectron *without Covalent*, then `dispatch_id` and `electron_node_id` will remain `None`, which fails the type check versus `int` and `str` respectively.

There a number of fields apart from these that have default values `None`, without `Optional[]` wrapping the type hint.

araghukas: if it's not clear from the preview above, I'm referring to `CircuitInfo` specifically

kessler-frost: Yeah, I agree there were several fields that were `None` and should have the `Optional` type there. I initially only fixed the ones which were causing issues, but ideally yes, we should fix all of the "None" ones to include `Optional` in them.

araghukas: Do you mind if I push a commit to fix this?

kessler-frost: Oh please go ahead that'll be quite helpful 🙏

PR #1757, opened 2023-08-21
Temporarily running nightly every hour now to fix

kessler-frost:

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -22,7 +22,7 @@ name: nightly
 
 on:
   schedule:
-    - cron: "0 4 * * *"
+    - cron: "0 * * * *"
 
 jobs:
   license:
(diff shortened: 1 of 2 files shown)


AmmarAgnostiq (approved):
I am approving but would like to know the reasoning on why it would succeed on an hourly basis rather than daily. Thanks.

AlejandroEsquivel (dismissed):
LGTM thanks @kessler-frost

AlejandroEsquivel:
@kessler-frost just noticed the seed data was changed and is causing one of the tests to fail, was the changes to the seed data intentional here ?

kessler-frost:
@AmmarAgnostiq oh, it's more so that if it fails, it'll fail quicker and we'll get to see why it did so. @AlejandroEsquivel I did change time for some of the seed data because they're missing a 0 in the nightly: https://github.com/AgnostiqHQ/covalent/actions/runs/5921818377/job/16054901914

kessler-frost:
It's failing for another one now, so I guess I'll have to add a 0 in a more "universal" way cuz of the assertion error. Could you take a quick look, I think there might be a better solution here rather than just adding a 0?

AlejandroEsquivel (approved):

PR #1764, opened 2023-08-31
Update qelectron_utils module for remote QElectron execution

araghukas:
Related PR for AWSBatch Executor: https://github.com/AgnostiqHQ/covalent-awsbatch-plugin/pull/79

Because the QServer is a fully-local object here, remotely executed QElectrons end up creating database entries (inside the `"qelectron_db_dir"`) *on the remote*. This means that database will be missing on covalent server, and the QElectron UI elements will therefore not work.

What we do here is to create the `"qelectron_db_dir"` (if it does not exist) when creating the `"results_dir"` entry. The former is for the QServer (and `Database`, and therefore the UI), while the latter is for the Covalent Server.

### DIFF
--- a/covalent/_shared_files/qelectron_utils.py
+++ b/covalent/_shared_files/qelectron_utils.py
@@ -98,7 +98,7 @@ def extract_qelectron_db(s: Union[None, str]) -> Tuple[str, bytes]:
     data_pattern = f".*{_QE_DB_DATA_MARKER}(.*){_QE_DB_DATA_MARKER}.*"
     if not s or not (_matches := re.findall(data_pattern, s)):
         app_log.debug(f"No Qelectron data detected in '{s!s}'")
-        return s, b''
+        return s, b""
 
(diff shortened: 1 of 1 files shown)


araghukas:
@cjao It looks the like the electron data layer [does in fact use a default `Database()` instance](https://github.com/AgnostiqHQ/covalent/blob/7c0cd147efa405bc26119fca436c30ea73c5f089/covalent_ui/api/v1/data_layer/electron_dal.py#L51). This ends up [looking for the QServers data store](https://github.com/AgnostiqHQ/covalent/blob/7c0cd147efa405bc26119fca436c30ea73c5f089/covalent/quantum/qserver/database.py#L51).

I would prefer a PR to fix the electron_dal over merging this PR, now that I think about it...

araghukas:
Losing this PR. Alternate, better fix found. https://github.com/AgnostiqHQ/covalent/pull/1766

PR #1766, opened 2023-09-01
Modify `Electrons` to use Covalent server's QElectron DBs, instead of QServer QElectron DB

araghukas:
The main issue here is that `Database()` by default points to data inside `~/.local/share/covalent/qelectron_db`. This folder belongs to the QServer.

We should instead use the transferred data in `~/.local/share/covalent/data/<dispatch_id>`, **because the former will NOT exist if the QElectron-containing electron is executed non-locally.**

**Solution:**

With the changes here, we still use the `Database` class to interface with the `.mdb` files. However, we don't have a *single* instance (no more `self.qdb`) that accesses an *single* directory (by default the undesired `~/.local/share/covalent/qelectron_db`).

Instead we specify the directory (i.e. `~/.local/share/covalent/data/<dispatch_id>/...`) every time we need to access *a* (one of arbitrarily many) QElectron database in the Covalent server's data folder.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+### Changed
+
+- Electron DAL to use Covalent server's data instead of QServer's data.
+
 ### Added
 
 - Corrected support from distributed Hamiltonian expval calculations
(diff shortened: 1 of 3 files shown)


araghukas:
@Prasy12 Just changed how the QElectron `Database` class is handled here. Nothing else should be affected.

cjao (approved):
Looks good @araghukas

PR #1767, opened 2023-09-02
Update `requirements.txt` to update pydantic minimum version and put `py.typed` in `MANIFEST.in`

### LINKED ISSUE #1765: Minimum version of Pydantic needs to be updated in `setup.py` to allow UI to work properly
### Environment

- **Covalent version**: 0.227.0-rc0
- **Python version**: 3.10.12
- **Operating system**: Ubuntu

### What is happening?

The UI is not showing the status of launched jobs.

image

Run an example Covalent script. I used my typical example below.

```python
import covalent as ct

@ct.electron
def add(a, b):
	return a+b

@ct.electron
def mult(a, b):
	return a*b

@ct.electron
def make_more(a):
	return [a]*3

@ct.electron
@ct.lattice
def add_distributed(vals):
	return [add(val,2) for val in vals]

@ct.lattice
def workflow(a, b, c):
	output1 = add(a,b)
	output2 = mult(output1,c)
	output3 = make_more(output2)
	output4 = add_distributed(output3)
	return output4

dispatch_id=ct.dispatch(workflow)(1,2,3)
ct.get_result(dispatch_id,wait=True)
```

### What should happen?

The UI should show the results.

### Any suggestions?

No :(

I tried completely removing my entire Conda env, removing the following files, and a few other things without luck. Reverting to 0.226.0-rc0 resolves the issue for me.

Directories removed: 
- `~/.cache/covalent`
- `~/.config/covalent`
- `~/.local/.share/covalent`

Strangely, just before this error I had a separate bug where the function names and descriptions for sublattice electrons were showing up under the wrong electrons. But I'll see if I can reproduce that error on >=0.227.0-rc0 once the UI is working.

arosen93:
- [] I have added the tests to cover my changes.

Closes #1765. Essentially, the existing minimum version for Pydantic causes the UI not to render appropriately. Pydantic > 2 is needed now for Covalent. As such, I have set the minimum version to 2.1.1. I chose 2.1.1 intentionally because: a) I tested it and it works; b) FastAPI strictly does not support 2.0.0, 2.0.1, or 2.1.0 so we can't use 2.0 through 2.1.0 anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,14 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+### Added
+
+- Added a `py.typed` file to support type-checking
 
 ### Changed
 
 - Raised the minimum version of Pydantic from 1.10.1 to 2.1.1 in `requirements.txt`
-
-### Added
-
-- Added a `py.typed` file to support type-checking
 
 ## [0.228.0-rc.0] - 2023-08-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+### Changed
+
+- Raised the minimum version of Pydantic from 1.10.1 to 2.1.1 in `requirements.txt`
+
 ## [0.228.0-rc.0] - 2023-08-31
 
 ### Authors
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+
 ### Changed
 
 - Raised the minimum version of Pydantic from 1.10.1 to 2.1.1 in `requirements.txt`
 
-## Added
+### Added
 
 - Added a `py.typed` file to support type-checking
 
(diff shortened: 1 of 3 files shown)


santoshkumarradha:
@arosen93 thanks a lot for identifying the exact cause ! This could have been quite nasty to debug sans UI logs ! 


@kessler-frost  can you take a look at this please 🙏🏻

kessler-frost:
Upgrading Pydantic might break certain things I believe (hopefully I'm wrong), but let's see what happens. This PR looks gtg from my side.

arosen93:
🙏

arosen93:
I added the `py.typed` file to `MANIFEST.in` as well here.

kessler-frost (approved):

wjcunningham7:
Just a heads up to the team Pydantic 2 has been known to not work on AWS Lambda, so keep an eye out for this during the next round of QA
https://twitter.com/pydantic/status/1678298418983124992
https://stackoverflow.com/questions/76650856/no-module-named-pydantic-core-pydantic-core-in-aws-lambda-though-library-is-i

CC @AlejandroEsquivel

wjcunningham7 (approved):

Prasy12:
There also is a case where pydantic 2 has been supported only for fastapi versions > 100. So we might need to tweak the fastapi limits in the requirements as well. (Currently <100 also can be installed)

Also when we install fastapi, pydantic gets installed automatically with a fastapi compatible version. So we might need to set both the fastapi and pydantic version correctly to avoid any failures.

cc : @kessler-frost @arosen93 @wjcunningham7

PR #1768, opened 2023-09-02
Add `py.typed` support for type-hinting

arosen93:

As noted in [PEP 561](https://peps.python.org/pep-0561/), a `py.typed` file is needed in order for others to use type-checking with the package, like with mypy. I added that here.


santoshkumarradha (approved):
Perfect !

santoshkumarradha:
@kessler-frost seems like all tests passed in this PR but merged develop is failing in.

arosen93:
There's never too simple a PR to break things it seems. 😅

arosen93:
There's never too simple a PR to break things it seems. 😅

kessler-frost:
>There's never too simple a PR to break things it seems. 😅

Usually it's because there are certain tests which are not run on PR creations in order to facilitate faster development. And when they are run, say during the creation of a pre-release or merge to develop, they show what is broken if Covalent was used as a package from PyPi as `pip install covalent` instead of the development mode of `pip install -e .` 😄 .

>@kessler-frost seems like all tests passed in this PR but merged develop is failing in.

I'm looking into it, it's the validate distribution test that's failing. So any new non-python related file added to the repo - if it also needs to be added to the packaged tar, will either need to be ignored (which isn't the case for this one), or added to the `MANIFEST.in`, similar to `requirements.txt`. That said, we should've also mentioned it in the contributing guide where it doesn't seem to have a mention of.

arosen93:
Of course, that makes perfect sense. Don't worry -- it was not a critique of you all :)

Ah okay, that makes sense re: the distribution. I am not 100% sure where it should be added (`mypy` was just complaining to me that Covalent didn't have it, so that's why I added it).

kessler-frost:
>Don't worry -- it was not a critique of you all :)

Haha, thanks 😄 ! But please don't hesitate to critique tho, it's part of open source development 😉 . I think you've added it in the right place, it just needs to be included in the `MANIFEST.in`.

arosen93:
>  it just needs to be included in the MANIFEST.in.

Makes sense! I think I saw the `recursive-include` commands in the `MANIFEST.in` file and assumed that it covered the `covalent` directory. I see that was not the case now!

PR #1778, opened 2023-09-15
Quick fixes for qelectron if found any

kessler-frost:


kessler-frost:
I should be able to fix the coverage for this.

AlejandroEsquivel (approved):

PR #1780, opened 2023-09-19
Exclude `aiohttp` 4.0.0a* in `requirements.txt`

Andrew-S-Rosen:

For some reason that I have yet to figure out, when I try to do `pip install --no-cache-dir covalent --pre` in a fresh Python 3.10 environment on Linux, it tries to install aiohttp 4.0.0.a1. However, 4.0.0a0 and 4.0.0.a1 are old pre-releases (from 2019) that have broken wheels. In this PR, I excluded these versions from `requirements.txt`.

>  WARNING: Requested aiohttp>=3.8.1 from https://files.pythonhosted.org/packages/f0/f2/703dba52c7620199ea0ec8ea9a4a2f06203b4893b94f60240c2c10225043/aiohttp-4.0.0a1.tar.gz (from covalent), but installing version 4.0.0a1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Excluded 4.0.0a* versions of aiohttp in `requirements.txt` 
 - Raised the minimum version of Pydantic from 1.10.1 to 2.1.1 in `requirements.txt`
 - Electron DAL to use Covalent server's data instead of QServer's data.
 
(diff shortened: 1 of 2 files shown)


wjcunningham7:
@Andrew-S-Rosen it appears `pip install X --pre` will target pre-releases not just for package `X` but also for anything it installs as a dependency.  In my opinion this is not the behavior we want.  You can instead do this:
```
pip install "covalent>=0.220.0rc0"
```
which would give you any pre-release after the latest stable release (0.220.0)

CC @AlejandroEsquivel @santoshkumarradha

Andrew-S-Rosen:
Oh, I had no idea! Well, that certainly clarifies things. I'll go ahead and close this PR then since it's a "me" problem. 😅

wjcunningham7:
> Oh, I had no idea! Well, that certainly clarifies things. I'll go ahead and close this PR then since it's a "me" problem. 😅

This was a valuable lesson, I wasn't aware of the behavior before

PR #1794, opened 2023-09-27
Adding qelectron to stable covalent

kessler-frost:


wjcunningham7 (approved):

Andrew-S-Rosen:
@kessler-frost: This PR introduced several strict version pins, which can cause challenges for developers planning to use Covalent in their own packages. Are all of these version pins strictly necessary? We should relax as many as possible and replace with `>=`. Also, just to confirm, `pennylane` should be a core requirement and not something to be installed as an "extra"?

kessler-frost:
@Andrew-S-Rosen : This was merged to one of the older stable branches where most of the things are working as expected. In `develop` the relaxed versions are already there and we will soon have a stable release based on our current `develop`.

The changes made in this PR and subsequently the place where this was merged to, are not to be considered as newer (or more up to date versions). They are just adding a new pretty much independent feature to an older covalent version.

Andrew-S-Rosen:
@kessler-frost: Thanks for clarifying! However, when I look at `requirements.txt` in `develop`, it has the strict version pins still: https://github.com/AgnostiqHQ/covalent/blob/develop/requirements.txt. Should I be looking somewhere else?

kessler-frost:
It has the strict ones for some of them (such as `lmdbm`, `pennylane`) as when tested we saw things to be only working on those. However, I see that you have a PR open for unlocking these versions, if things don't fail now there then let's relax those restrictions.

Andrew-S-Rosen:
@kessler-frost: Thanks! The one I was really curious about was `orjson`. So, I'll iteratively try things and see how it fares. I decided to start with unpinning them all. Then I'll revised as-needed.

PR #1840, opened 2023-10-23
hacktoberfest - Added a get all results function

### LINKED ISSUE #1701: Adding a `ct.get_all_results()` function
### What should we add?

There is currently a `ct.get_result(<dispatch ID>)` function, which works perfectly. It would be even more awesome if there were an option to do `ct.get_all_results()` that would automatically return all results!

The "manual" way I'd come up with is something like this:
```python
config_results_dir = ct.get_config()["dispatcher"]["results_dir"]
dispatch_ids = os.listdir(config_results_dir)

# Populate the docs
docs = []
for d_id in dispatch_ids:
    try:
        result_obj = ct.get_result(d_id)
    except MissingLatticeRecordError:
        continue
    if result_obj and result_obj.status == "COMPLETED":
        docs.append({"dispatch_id": d_id, "result": result_obj.result})
return docs
```

I'm not sure if there is a more efficient way. Of course, I'm happy to open a PR based on this if what I've proposed is the optimal route.

Looping, like above! :)

Smartmind12:
The get_all_results() function to retrieve all results, optionally filtered by a list of dispatch IDs, statuses, and timestamps. This provides more flexibility and control over the results you want to retrieve.

Could potentially close #1701 

### DIFF
--- a/covalent/_results_manager/results_manager.py
+++ b/covalent/_results_manager/results_manager.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 import contextlib
 import os
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import Dict, List, Union, Optional
 
 from furl import furl
 from requests.adapters import HTTPAdapter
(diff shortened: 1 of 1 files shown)


Smartmind12:
Hey @Andrew-S-Rosen Kindly review this PR that closes the above issue!

wjcunningham7:
Hi @Smartmind12 thanks for the PR.  There are few changes needed before we can accept it.

In your implementation you assume the Covalent server is running locally and directly interact with the results directory.  This pattern is discouraged, in favor of an API-based approach.  The client method in the Covalent SDK (`/covalent`) should call an API defined in the Covalent Dispatcher (`/covalent_dispatcher/_service/app.py`).  Further, you'll want to apply filtering on the server side rather than the client side.  This amounts to adding a batch-get version of the route `/api/v1/dispatches/:dispatch_id`, such as `/api/v1/dispatches` where filters are applied as query strings.  Finally, any batch-get response should include pagination.

An example of a similar route can also be found in the front-end code, since the Covalent dashboard also performs a batch-get of dispatches (see /covalent_ui/api/v1/routes/end_points/summary_routes.py`)

wjcunningham7 (changes requested):
See other comment

adirao-projects (changes requested):
I'm not sure if ```ct``` would be defined inside this function as ```ct``` (I'm assuming) is the covalent package itself. Different implementation is required.

FyzHsn:
There hasn't been any activity in over a year. Going to close this one.

PR #1849, opened 2023-11-14
Fix propagation of lattice-default attributes to electrons

### LINKED ISSUE #1781: Error when setting `executor` in `@ct.lattice`: `DaskExecutor` object has no attribute `scheduler_address`
### Environment

- **Covalent version**: >=0.227.0rc0
- **Python version**: 3.10.13
- **Operating system**: Linux

### What is happening?

When setting the `executor` kwarg in the `@ct.lattice` decorator, I receive the following error message in the post-processing step of the workflow.

```bas
'DaskExecutor' object has no attribute 'scheduler_address'
```

**Note**: setting the `executor` kwarg in the `@ct.electron` (but not `@ct.lattice`) works fine and the workflow goes to completion. This seems very closely related to #1733, which is perhaps not fully solved.

image

```python
import covalent as ct

executor = ct.executor.LocalExecutor()

@ct.electron
def add(a, b):
    return a + b

@ct.lattice(executor=executor)
def workflow(a, b):
    return add(a, b)


dispatch_id = ct.dispatch(workflow)(1, 2)
result = ct.get_result(dispatch_id,wait=True)
assert result.status == "COMPLETED"
```

### What should happen?

The workflow post-processing step should complete.

### Any suggestions?

I need to see if the same behavior occurs on `main`.

cjao:
Also use `workflow_executor` to build sublattice graphs

Closes #1781 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Lattice-default metadata attributes are now applied correctly
+- Sublattices are built using `workflow_executor`
 - Added covalent version attribute to Remote Executors
 - Removed unassigned variable names
 - Contributing guidelines steps for installing for the first time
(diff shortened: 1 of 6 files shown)


Andrew-S-Rosen:
Huge thank you for this!!

kessler-frost (approved):

PR #1850, opened 2023-11-14
Post mem improvement Qelectron fixes

kessler-frost:


kessler-frost:
It's working as expected, now need to fix the tests.

kessler-frost:
- [ ] Remove the `qelectron_db_filename` variable if possible since it is always going to be "data.mdb", also since only the db path is required for `Database` class. Make sure it doesn't interfere when downloading the asset to object store - in which case we should let it be here.
- [ ] Don't load the `qelectron_db` unconditionally when doing `get_result`, i.e. add the key to the `DEFERRED_KEYS` variable so that it only gets loaded when it's actually needed
- [ ] Remove `value` from the `DEFERRED_KEYS` variable

AlejandroEsquivel (approved):

PR #1862, opened 2023-11-23
Temporary more frequent nightlies

kessler-frost:


kessler-frost:
So, @AlejandroEsquivel I wanted this change because it is easier for me to stop the nightly workflow than to trigger it manually, which is why it would be helpful to have it this frequent. I'll keep it disabled and only enable it once I think it is going to work. This would help tremendously for debugging temporarily.

AlejandroEsquivel (approved):

PR #1863, opened 2023-11-24
Fixing nightly

### LINKED ISSUE #1839: Fixing tests which are failing on nightly
The last run nightly showed that there are several separate tests which are failing on separate OS/Python version combinations: https://github.com/AgnostiqHQ/covalent/actions/runs/6581710795. This should be fixed so that we can have a pre-release soon as well as the `master` gets updated with the new license.

kessler-frost:

Resolves #1839

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -20,6 +20,10 @@ on:
   schedule:
     - cron: "*/10 * * * *"
 
+permissions:
+  id-token: write
+  contents: read
+
 jobs:
   license:
     name: License Scanner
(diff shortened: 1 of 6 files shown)


kessler-frost:
Removing CentOS support makes sense at this point due to its end of life: https://www.centos.org/centos-linux-eol/, CentOS Linux project has been phased out essentially and the move is being made towards CentOS Stream.

kessler-frost:
Turns out Github has removed [support for node 16](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/) and node 20 requires certain glib libraries which are not present in centOS 7 which is why the `checkout` action fails. Thus, we had to remove centOS 7 from the test matrix.

wjcunningham7 (approved):

Issue #1845, opened 2023-11-09
AttributeError: type object 'DefaultQubit' has no attribute 'operations'


kessler-frost:
Recently, Pennylane has removed the `operations` attribute from their `DefaultQubit` class which means the following error might be thrown for users installing covalent in a new environment with the pennylane version > 0.31.1 - we've observed 0.31.1 doesn't have this issue:

```python
AttributeError: type object 'DefaultQubit' has no attribute 'operations'
```

This can be solved by installing `pennylane==0.31.1`, but for now we'll need to restrict the limit of pennylane from >=0.31.1 to ==0.31.1 until we switch to the new method, so that it doesn't break covalent.

PR #1864, opened 2023-11-25
Change `pennylane==0.33.1` to `pennylane>=0.31.1,<0.33.0` and remove redundant version from `tests/requirements.txt`

Andrew-S-Rosen:

I removed the strict version pin on pennylane. It was forcing an upper-bound for the numpy version.

I saw that in #1845, it was mentioned that there can be issues installing Covalent with the most recent versions of pennylane. However, the test suite passes, so I'm curious where the error resides. If there is indeed an error, then something should be added to the test suite to flag it.

Sorry --- I'll continue to be annoying about fixing a dependency to a single version in setup.py unless it's critical. 😉


kessler-frost:
Hi @Andrew-S-Rosen ! Thanks a lot for this 😄 , this exposed that we haven't removed the pennylane requirement from `tests/requirements.txt` which installs 0.31.1 after the normal `requirements.txt` installs 0.33.1 - which is why the tests don't fail. Try removing that line from `tests/requirement.txt` in this PR?

kessler-frost:
Although, I do agree that a better solution would be to put a range on the `requirements.txt`, seems like `>=0.31.1,<0.33.0` should work.

Andrew-S-Rosen:
@kessler-frost: Ah, that explains some things! You're right, the test suite fails (as expected) when removing the `pennylane` requirement in the `tests` folder.

So, for this PR, I have done the following:
- Removed `pennylane` from `tests/requirements.txt` so it's only in one plalce
- Changed it from a strict pin to 0.33.1 to a range from >=0.33.1,<0.33.1

kessler-frost:
Thanks @Andrew-S-Rosen ! I think you've mentioned the version range wrong 😅 `pennylane>=0.33.1,<0.33.1` instead of `pennylane>=0.31.1,<0.33.0` (in the changelog as well).

Andrew-S-Rosen:
@kessler-frost: Looks like I need to go back to middle school algebra!!

kessler-frost:
Also, in the requirements.txt file the upper limit is 0.33.1 instead of 0.33.0 :P, which installs 0.33.0 and the test fails because of that 😅 .

kessler-frost:
I'm sorry I should've reviewed it all in one go.

Andrew-S-Rosen:
@kessler-frost not to worry, I realized the same. This is what happens when I open a PR while distracted. 😅 Sorry about that.

kessler-frost (approved):

santoshkumarradha (approved):

PR #1876, opened 2023-12-06
Adding Module Dependencies feature

kessler-frost:


santoshkumarradha:
@kessler-frost reminder on the usual list of docs changes for this as well along with this PR

CC: @sriranjanivenkatesan once the docs are up, please keep a tab to update it

kessler-frost:
One thing to note here is that DepsModule are kind of a fake traditional “deps” in the sense that we don’t require them to be passed to the server as cloudpickle takes care of serializing the module when the serialization of the function happens.

kessler-frost:
One drawback of this is that there we won't be recording the name of the module in the dispatcher server. @santoshkumarradha it's a UX question as in do we want to have a record or not? This is mostly an additive change and can be taken care of in a separate PR as well.

santoshkumarradha (approved):
UX is gtg 🤘

cjao (approved):

PR #1879, opened 2023-12-08
Reduce the number of asset transfers during dispatch (1/2): omit null assets

### LINKED ISSUE #1881: Reduce asset uploads during dispatch (1/2): omit null assets
(wip)

Electron attributes unset during dispatch submission are currently
initialized to some concrete default value of the expected asset type
(empty strings for text assets, pickled `None` values for electron
outputs) and uploaded individually to the dispatcher. These uploads are
superfluous as the attributes will be overwritten when the workflow
runs.

We can avoid such uploads by adhering to the following rules:

-   Any electron, lattice, or result attribute is assigned `None` &#x2013; the unique value of `NoneType` &#x2013; if and only if it is unset. Note that `None` should be distinguished from the nil value of other types (empty strings, `pickle.dumps(None)`).

-   Do not transfer any unset attributes when submitting a dispatch or retrieving results.
-   An asset record in the server's DB has `size = 0` precisely if it is not backed by a file

file in `$COVALENT_DATA_DIR`.

The concrete changes required are:

- [ ] All unset Electron, Lattice, and Result attributes should be   initialized to `None`.
- [ ]  When serializing the workflow for submitting to the dispatcher,   represent each `None` attribute as an asset with `size==0`.
- [ ]  When uploading assets during dispatch submission, skip any assets with `size==0`. 
- [ ]  When retrieving results, don't download any assets with zero size,   and simply interpret their corresponding attribute values as `None`.
- [ ]  Ensure that asset sizes are updated in the DB whenever their data in   the filesystem changes. In particular, executor plugins using the new API should always return the size    (number of bytes) of each (serialized) task artifact in their  `receive()` methods.

cjao:
Initialize unset assets to `None` and don' upload them.

When downloading assets during `get_result()`, ignore assets with zero size.

Also require asset sizes when retrieving asset updates in `Executor.receive()`.

Closes #1881 

### DIFF
--- a/covalent/_shared_files/schemas/asset.py
+++ b/covalent/_shared_files/schemas/asset.py
@@ -28,11 +28,13 @@ class AssetSchema(BaseModel):
     remote_uri: Optional[str] = None
 
     # Size of the asset in bytes
-    size: Optional[int] = 0
+    size: int
 
 
 class AssetUpdate(BaseModel):
     remote_uri: Optional[str] = None
-    size: Optional[int] = None
     digest_alg: Optional[str] = None
     digest: Optional[str] = None
+
+    # Size of the asset in bytes
+    size: int
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [UNRELEASED]
 
+### Changed
+
+- SDK no longer uploads empty assets when submitting a dispatch.
+- Results Manager avoids downloading assets with size 0.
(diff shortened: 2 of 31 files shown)


covalent/_shared_files/schemas/asset.py
```diff
@@ -28,11 +28,13 @@ class AssetSchema(BaseModel):
     remote_uri: Optional[str] = None
 
     # Size of the asset in bytes
-    size: Optional[int] = 0
+    size: int
```
FyzHsn, line 31: I see that size is not optional anymore. Just wanted to double check that you did a grep to check if all the sizes are set.

FyzHsn (approved):

PR #1883, opened 2023-12-11
Reduce memory overhead for TransportableObject

cjao:
* Always represent TransportableObject internally as a single array of bytes. Various properties, such as `header`, or `object_string`, decode various segments of the byte array.

* Store the serialized object as raw picklebytes without base64-encoding. As a result, `get_deserialized()` no longer needs to create a temporary copy of the raw picklebytes. The data segment is directly unpickled. Base64-encoding is applied to the data segment or the entire internal buffer whenever a print friendly representation of the `TransportableObject` is desired.

* Since the properties of `TransportableObject` are simply views into the underlying buffer, `TransportableObject` may itself be serialized efficiently by simply writing out the byte array.
`


cjao:
Tests pass but codecov fails.

cjao:
Gentle reminder @wjcunningham7 @FyzHsn : )

PR #1890, opened 2023-12-16
Assets: support range requests for alt representations

cjao:


cjao:
The changes here allow one to download a specific byte range of an asset's string representation. This may be useful for rendering large assets in the UI.

PR #1901, opened 2024-01-07
Adding a ct.get_all_results() implementation 

adirao-projects:
Added an implementation of the ```get_all_results``` function desired in #1701 . The main issue with the current implementation I've added is that it requires either of the following.

1. The user to provide a list of dispatch ids
2. The user to provide a dispatcher and results directory and be running on their own machine

If these requirements are not met, the provided implementation will not work. The internal documentation I've provided explains the logic on how everything works. I have not yet updated the external documentation. No tests have been written to test this function yet.


adirao-projects (commented):
Seems like this is the same as what was there before, but the bot thinks it's better format wise

FyzHsn:
Outdated

PR #1916, opened 2024-01-22
Enabling qelectron tests and making them opt-outable

kessler-frost:

This PR makes the following changes:
1. Adds the qelectron tests to the `tests` github workflow
2. Makes qelectrons a feature that can be opted out of from since it might be blocking some users from upgrading certain versions even if they do not require those in their workflows
3. Suppresses the harmless sqlalchemy warning that shows up on a fresh covalent start


kessler-frost:
The diff coverage is low because the functions that have been moved to another place didn't already have tests in place.

kessler-frost:
@Andrew-S-Rosen I think once this PR gets merged, it should unblock you in the sense that we won't have a numpy dependency for covalent. The only thing needed when installing covalent would be to have an environment variable set i.e. `COVALENT_DISABLE_QELECTRON=1`, and none of the quantum stuff will be installed - thus neither will numpy so you can have whichever numpy version you need.

Simple install command to install the pre-release without numpy dependency once this PR gets merged:
```bash
COVALENT_DISABLE_QELECTRON=1 pip install --pre covalent
```

Andrew-S-Rosen:
Brilliant, thank you!

kessler-frost:
To make it further easier, now qelectrons (and its dependencies) are an opt-in only feature (instead of opt-out). In case its installation is required by someone, they will be able to install it using `pip install covalent[quantum]`. By default none of the quantum dependencies will be installed.

cjao (approved):
Looks good @kessler-frost

PR #1924, opened 2024-02-01
Executor improvements

mpvgithub:
### Changed

- Modified the executor core to change the file path for terraform state files to store and read from **.config/executor_plugins** folder
- Clean up any half-done/dirty deployed resources post **deploy up**

### DIFF
--- a/covalent/cloud_resource_manager/core.py
+++ b/covalent/cloud_resource_manager/core.py
@@ -27,6 +27,7 @@ from types import ModuleType
 from typing import Callable, Dict, List, Optional, Union
 
 from covalent._shared_files.config import set_config
+from covalent._shared_files.defaults import get_default_sdk_config
 from covalent.executor import _executor_manager
 
 logger = logging.getLogger()
@@ -34,6 +35,7 @@ logger.setLevel(logging.ERROR)
 handler = logging.StreamHandler(sys.stderr)
 logger.addHandler(handler)
 logger.propagate = False
+sdk_constants = get_default_sdk_config()
 
 
 def get_executor_module(executor_name: str) -> ModuleType:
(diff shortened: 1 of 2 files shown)


covalent/cloud_resource_manager/core.py
```diff
@@ -27,13 +27,15 @@
...
 from covalent._shared_files.config import set_config
+from covalent._shared_files.defaults import get_default_sdk_config
 from covalent.executor import _executor_manager
 
 logger = logging.getLogger()
 logger.setLevel(logging.ERROR)
 handler = logging.StreamHandler(sys.stderr)
 logger.addHandler(handler)
 logger.propagate = False
+sdk_constants = get_default_sdk_config()
```
kessler-frost, line 38: We should use `get_config` instead of directly importing the defaults since they might've been changed by the user in the config file directly. Using `get_config` with specifically what is needed (instead of the entire sdk config) will take that into consideration as well.

covalent/cloud_resource_manager/core.py
```diff
@@ -389,7 +397,24 @@ def _get_tf_statefile_path(self) -> str:
 
         """
         # Saving in a directory which doesn't get deleted on purge
-        return str(Path(self.executor_tf_path) / "terraform.tfstate")
+        return str(Path(self.executor_tfstate_path) / "terraform.tfstate")
```
kessler-frost, line 400: >Saving in a directory which doesn't get deleted on purge

`self.executor_tfstate_path` seems to point to `~/.config/covalent/executor_plugins/*`, this WILL get deleted by purge. We need to add an exception in the purge command to prevent deletion of these state files.

covalent/cloud_resource_manager/core.py
```diff
@@ -389,7 +397,24 @@ def _get_tf_statefile_path(self) -> str:
...
+            None
+
+        Returns:
+            Path to terraform state file
+
+        """
+        # Saving in a directory which doesn't get deleted on purge
+        state_path = os.path.join(parent_path, self.executor_name)
+        if not os.path.exists(state_path):
+            os.makedirs(state_path)
```
kessler-frost, line 416: Same comment as above. One minor thing is that if possible let's try to use `pathlib`'s `Path` instead of `os.path` to join/checking existence/checking whether its a directory/etc.

kessler-frost (changes requested):

PR #1926, opened 2024-02-06
Parse electron function strings at Electron construction

cjao:
Electron function strings were previously parsed during `build_graph()`. This doesn't work for electrons in sublattices since they will be built in an executor which typically won't have access to the user's source tree.

With this change, all function strings will be parsed client-side when the electron decorator is evaluated.

Tested manually using both Dask and a modified AWSBatch executor image.


santoshkumarradha (approved):
@cjao non blocking, but might be better to have sublattice string test as well along with this?

PR #1928, opened 2024-02-08
Removing my email from the CHANGELOG

Andrew-S-Rosen:
This PR is to remove my email from the `CHANGELOG` and to replace it with my GitHub username.

When searching for my email via Google, the [only hit](https://www.google.com/search?q=%22asrosen93%40gmail.com%22) is for the Covalent `CHANGELOG`. I've been getting a lot of spam in the last several days, so I would like to clean up this public trace of my email just in case. I have replaced my email with my GitHub username.

By default, it may be worth considering using usernames by default in the `CHANGELOG`. In any case, I'm aware that this is an automated process so this PR is only a band-aid solution. I wasn't sure where to adjust the automated procedure. Also, this PR will obviously fail for updating the `CHANGELOG`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Authors
 
-- Andrew S. Rosen <asrosen93@gmail.com>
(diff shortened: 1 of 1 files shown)


Andrew-S-Rosen:
Thanks, @kessler-frost!

PR #1929, opened 2024-02-08
Removing author email from changelog

kessler-frost:


AlejandroEsquivel (changes requested):
change looks good however I think co-authors will still have emails
might be just having:
return `- ${author.split(/[<>]/)[1].trim()}\n`;  
in line 16

AlejandroEsquivel (approved):
Looks good @kessler-frost , only thing i'm not so sure about is if `!authors.includes(author.email)` was doing any deduplication, but let's give it a shot!

Issue #1539, opened 2023-03-03
Inevitable recursion error from inside get_result


araghukas:
### Environment

- **Covalent version**: 0.209.1
- **Python version**: 3.8
- **Operating system**: MacOS

### What is happening?

If `ct.get_result(dispatch_id, wait=True)` takes long enough, eventually a `RecursionError` will be raised.

```
Traceback (most recent call last):
  File "/Users/araghukas/miniconda3_x86_64/envs/mkl-rosetta/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/araghukas/miniconda3_x86_64/envs/mkl-rosetta/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/araghukas/Code/algo-multikernel/software/scripts/qcc_net/__main__.py", line 289, in <module>
    fire.Fire(QCCExperiment, name="qcc_net")
  File "/Users/araghukas/miniconda3_x86_64/envs/mkl-rosetta/lib/python3.8/site-packages/fire/core.py", line 141, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
  File "/Users/araghukas/miniconda3_x86_64/envs/mkl-rosetta/lib/python3.8/site-packages/fire/core.py", line 475, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
  File "/Users/araghukas/miniconda3_x86_64/envs/mkl-rosetta/lib/python3.8/site-packages/fire/core.py", line 691, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
  File "/Users/araghukas/Code/algo-multikernel/software/scripts/qcc_net/qcc_experiment.py", line 77, in run
    records = self._run(random_seeds, self.data_dimension)
  File "/Users/araghukas/Code/algo-multikernel/software/scripts/qcc_net/qcc_experiment.py", line 165, in _run
    return self._execute(kwargs)
  File "/Users/araghukas/Code/algo-multikernel/software/scripts/qcc_net/qcc_experiment.py", line 179, in _execute
    records = ct.get_result(dispatch_id, wait=True).result
  File "/Users/araghukas/miniconda3_x86_64/envs/mkl-rosetta/lib/python3.8/site-packages/covalent/_results_manager/results_manager.py", line 57, in get_result
    result = _get_result_from_dispatcher(
[...]

PR #1949, opened 2024-05-21
Fix `get_result(wait=True)` and support Python 3.12

### LINKED ISSUE #1973: Cannot import covalent with Python >=3.12
The `pkg_resources` module was [removed from the main Python distribution](https://docs.python.org/3/whatsnew/3.12.html) as of 3.12.

### Steps to reproduce
1. Create a new virtualenv `python3.13 -m venv venv`
2. Install Covalent in the virtualenv
3. Try to import covalent:
```
Python 3.13.1 (main, Dec  9 2024, 00:00:00) [GCC 14.2.1 20240912 (Red Hat 14.2.1-3)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import covalent
  File "<python-input-0>", line 1, in <module>
    import covalent
  File "/var/home/casey/programming/govalent/.venv/lib64/python3.13/site-packages/covalent/__init__.py", line 23, in <module>
    from . import executor, leptons  # nopycln: import
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/home/casey/programming/govalent/.venv/lib64/python3.13/site-packages/covalent/executor/__init__.py", line 30, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

For alternatives see the [migration guide](https://importlib-resources.readthedocs.io/en/latest/migration.html).

cjao:
`get_result(wait=True)` will still [error out after some time](https://github.com/AgnostiqHQ/covalent/blob/develop/covalent/_results_manager/results_manager.py#L518). Thi PR proposes a fix using a combination of API and SDK improvements.

The previous `GET /dispatches/{dispatch_id}` endpoint was trying to do too much. Its responsibilities are now separated into two endpoints:

* `GET /dispatches`: bulk query dispatch summaries (including status) with options to filter by `dispatch_id`, sort chronologically, and also limit the output to status only.

* `GET /dispatches/{dispatch_id}`: download manifest

To achieve the desired behavior of `get_result(id, wait=True)`, the client

1. Polls the dispatch status by querying the first endpoint.

2. Downloads the manifest after the dispatch has reached a final status.

The server no longer returns 503 errors when the dispatch is not yet "ready". A 503 status code is not appropriate here because it is intended to convey temporary service unavailablity resulting from server overload or rate limiting. However, the fact that the workflow is still running does not indicate any fault of the server.

These changes will allow `get_result(dispatch_id, wait=True)` to wait as long as required instead of erroring out after some time.

Fixes https://github.com/AgnostiqHQ/covalent/issues/1539
Fixes #1973 
Fixes ENG-2039


santoshkumarradha:
@cjao can you see if you make this true in a way for other waits as well like env and deployment ?

covalent_dispatcher/_dal/result.py
```diff
@@ -45,6 +46,41 @@
 class ResultMeta(Record[models.Lattice]):
     model = models.Lattice
 
+    @classmethod
+    def get_toplevel_dispatches(
```
FyzHsn, line 50: If this is not used, delete

FyzHsn (changes requested):

FyzHsn (approved):

wjcunningham7 (approved):

PR #1981, opened 2025-03-24
File transfer and sublattice enhancements

cjao:
Introduce file transfer and sublattice enhancements to decentralize asset handling

### File Transfers

Improve automatic file transfer strategy selection by automatically
handling the following cases:

- Local-Remote
- Remote-Local
- Local-Local

Allow registering new file transfer strategies to handle custom URL schemes

Enhance HTTP strategy to support uploads

Centralize and abstract file transfers in SDK. Dispatch submission and
retrieval now use FileTransfers to upload and download assets

Compute `Result.error` client side to avoid needless server-side writes
to object store

For asset uploads/downloads, decouple the steps of
1) Obtaining the upload/download URL
2) Performing the file transfer

### Sublattices

Add endpoint to link sublattice with parent. Link sublattice dispatch with parent electron in `_build_sublattice_graph`. The dedicated sublattice registration endpoint `POST
/dispatches/{dispatch_id}/sublattices` is no longer used.

Adjust server-side handling of sublattice dispatches in `update_node_result`. Since the sublattice `dispatch_id` is now linked to the parent electron during `_build_sublattice_graph`, the server no longer needs to read the sublattice manifest from object storage if it can see the link in the DB.

Improve handling of sublattices built using executors that can't reach the control plane (e.g. AWS Batch -|-> Covalent running on local workstation).
- When `_build_sublattice_graph` is unable to submit the sublattice directly, it tars up the staging directory returns the tarball to the control plane.
- Add a new importer to 1) extract the tarball, 2) import the sublattice manifest, 3) transfer assets to object storage.
- Replace legacy sublattice importer. Future improvements will reduce the server memory footprint from handling these sublattices (see code comments about streaming the tarball to the extractor)
- Remove obsolete endpoint `POST /dispatches/{dispatch_id}/sublattices`

**Misc improvements**

Sublattices: Only copy db references when copying assets

Sublattice result -> parent electron output
Final electron output -> Dispatch result
Original electron attributes -> redispatched electron attributes

Sublattice: Back up manifest asset record before overwriting it

Cleanup: remember to hold refs to background tasks

Cleanup: refactor URI filtering

Cleanup: remove unused APIs

- [x] I have added the tests to cover my changes.
- [x] I have updated the documentation and CHANGELOG accordingly.
- [x] I have read the CONTRIBUTING document.

PR #1982, opened 2025-03-26
Move in-memory dispatcher state to the main DB

cjao:
The dispatcher currently maintains a small amount of in-memory state consisting of several quantities that would be too costly (`O(num_electrons)`) to compute de novo from the database each time they are needed (`O(1)` times per electron). These are:
- Number of unresolved tasks in the dispatch (`num_submitted - num_teminal)`
- Number of dependencies of a task group that have yet to complete.
- The list of topologically sorted electrons within a task group.

Remembering these quantities over the course of a dispatch amortizes the initial computational cost over all of the electrons in the workflow.

This PR allows one to store the state in a real SQL DB by setting the connection string in the environment variable `COVALENT_CACHE_DB_URL`. By default, the cache db will live in an ephemeral sqlite DB that will be cleaned up when the server shuts down. If `/tmp` is a tmpfs mount like on many recent Linux distributions, the ephemeral DB will live purely in memory.

Benchmarks:

1000 trivial electrons, workflow processing time dominated by dispatcher I/O
```
import covalent as ct

@ct.electron
def task(x):
    return x + 1

@ct.lattice
def workflow(x):
    res = task(x)
    for _ in range(999):
        res = task(res)
    return res

dispatch_id = ct.dispatch(workflow)(0)
print(dispatch_id)
print(res)
```

With default settings (cache DB on tmpfs):
```
(venv) casey@datarobot covalent (bye-redis)$ time python benchmark.py 

real    1m46.539s
user    0m4.526s
sys     0m1.121s
```

With `COVALENT_CACHE_DB_URL` set to the connection string for the main dispatch DB (on SSD)
```
(venv) casey@datarobot covalent (bye-redis)$ time python benchmark.py 

real    2m55.202s
user    0m4.548s
sys     0m1.165s
```

Current develop (dispatcher state in python dictionary):
```
(venv) casey@datarobot covalent (develop)$ time python benchmark.py 

real    1m40.423s
user    0m4.521s
sys     0m1.117s
```


cjao:
The external dep `pytest-asyncio` is somehow failing the license check.

cjao:
This one is less time sensitive than #1981  and #1883  @wjcunningham7

PR #2014, opened 2026-04-13
Make `dask[distributed]` optional via `covalent[dask]`- no dask install by default

araghukas:
Require `pip install covalent[dask]` to install `dask[distriburted]` dependency for covalent, do not install `dask[distributed]` otherwise, with fallback to `"local"` executor.

#### Main point
- Make `dask[distributed]` and optional dependency via `covalent[dask]` extra
- Default executor changed from `dask` to `local` in `defaults.py`

#### Necessary fixes
- Dropped Python 3.10 support, added Python 3.13; updated `test_matrix.json` and CI workflows accordingly
- Fixed `extract_graph` to handle both `"edges"` (networkx ≥3.x) and `"links"` (older networkx) as the edge key; added lower bound on networkx version
- Fixed various 422 errors by setting additional `Content-Type: application/json` headers as necessary
- Moved `locust` out of `tests/requirements.txt` into the benchmark workflow only, to avoid a `gevent` conflict
- Various test fixes: renamed `status` → `job_status`, corrected mock paths, updated graph key references


araghukas:
closing to try to bypass now-removed checks on Python 3.10

araghukas:
replaced with: https://github.com/AgnostiqHQ/covalent/pull/2016

